Issue #13, opened 2022-09-21
Roadmap


johnkit:
Current efforts are directed as follows:

* Extend the current (Seb's) trame-based prototype viewer for more general datasets (integrating pyvista-xarray).
* Extend the viewer for variable resolution level, with an explicit control (slider) for now.
* Extend the viewer for time-setp selection, with an explicit control (slider).
* Secure datasets  for testing and demonstration, ideally covering post-processing and AI applications, in hdf5 and zarr formats.
* Investigate potential strategies for adapting and/or reworking trame to the Jupyter widget ecosystem.

Priorities are still in flux. More comprehensive planning is needed and _should_ be forthcoming.

Issue #12, opened 2022-09-07
Viewer capabilities


jourdain:
- [x] Load Zarr file
- [ ] Create Grid and link zarr array to grid
   - [x] Image Data
   - [x] Rectilinear grid
      - [x] Use array for xyz point
      - [x] Use array for xyz cells
   - [x] Add point data 
      - [ ] Support extra dimension for time
   - [x] Add cell data 
      - [ ] Support extra dimension for time
- [x] Viewer controls
   - [x] Color by Array
   - [x] Toggle edge visibility
   - [x] Reset camera
   - [ ]  Level of detail slider
 - [ ] Load HDF5 file
 - [ ] Jupyter Kernel level integration for restore and save

jourdain:
<img width="1594" alt="Screen Shot 2022-09-07 at 17 09 29" src="https://user-images.githubusercontent.com/294518/188999491-418bd556-b0d8-4fa7-933b-1a3b02fc72a5.png">
<img width="1594" alt="Screen Shot 2022-09-07 at 17 10 29" src="https://user-images.githubusercontent.com/294518/188999494-262fc47a-173f-4373-8490-ae185e11ca30.png">
<img width="1594" alt="Screen Shot 2022-09-07 at 17 12 49" src="https://user-images.githubusercontent.com/294518/188999498-298fc2b7-aee2-4bef-9d26-edee3c8918cd.png">

jourdain:
@johnkit please add item in the todo list

@aashish24 let's figure out what are the next priorities and timeline. So far I did the bare minimum with the time I had available. Maybe @banesullivan can push some of it forward. But for now, I won't be able to push anything more this week.

aashish24:
This looks great. I edited (if that was what you were asking) the to-do list at the top. I think the next step would be the LOD slider that changes the resolution of the Mesh. And then, we need to look into if we could do something similar for the HDf5 datasets.

banesullivan:
> Jupyter Kernel level integration for restore and save

This subtask is not feasible with trame at this time without major work to `wslink` and `trame`

PR #14, opened 2022-10-05
Refactoring viewer to focus on Xarray

banesullivan:
This refactors the viewer so that it is an xarray-focused 3D viewer instead of general Zarr data.

For #12 and #13

The viewer has users select a `DataArray` in an Xarray `DataSet`, then select the X, Y, Z, and Time coordinate arrays before generating a mesh that is shown in the viewer.

## To do

- [x] Add time slider
- [ ] Better handle errors if trying to build mesh when arrays are not selected
- [x] Add level of detail selection option
- [ ] Add more examples (perhaps a cloud-hosted dataset)
  - [ ] add some to DKC and verify things work via URLs
- [x] Add a URL field to specify the path/URL
- [x] Drop down menu to select sample datasets

### Nicities

- [x] Disable XYZT selections when array is not chosen
- [x] Add slider to scale Z axes (or all axes??)
- [ ] Improve the mesh updating/building so that it is more automatic
- [ ] Use GeoVista for global plotting (ref https://github.com/pyvista/pyvista-xarray/blob/main/examples/level_of_detail_geovista.py)

Requires:

- https://github.com/pyvista/pyvista-xarray/pull/39
- https://github.com/pyvista/pyvista/pull/3318


banesullivan:
Merging for deployment/demo. Further improvements will come later

PR #23, opened 2023-06-27
UI redesign

annehaley:
- Move dataset selection box to sidebar
- Automatically select first available array once a dataset is ready
- Remove "reset view" button, automatically invoke reset on state changes
- Add error banner above render area to tell the user when/why an auto view reset fails
- Reorganize axis selection options to reduce redundant information

### DIFF
--- a/pan3d/viewer/engine.py
+++ b/pan3d/viewer/engine.py
@@ -56,8 +56,8 @@ class MeshBuilder:
         ]
         self._state.coordinates = []
         self._state.dataset_ready = True
-        # Set first array as active
-        # TODO self._state.array_active = list(dataset.data_vars.keys())[0]
+        if len(self._state.data_vars) > 0:
+            self._state.array_active = self._state.data_vars[0]["name"]
 
     def clear_dataset(self, **kwargs):
         self._state.dataset_path = None
@@ -65,6 +65,7 @@ class MeshBuilder:
         self._state.coordinates = []
         self._state.dataset_ready = False
         self._state.array_active = None
+        self._state.active_tree_nodes = []
         self._state.grid_x_array = None
         self._state.grid_y_array = None
         self._state.grid_z_array = None
@@ -81,9 +82,11 @@ class MeshBuilder:
 
     def on_active_array(self, array_active, **kwargs):
         if array_active is None or not self._state.dataset_ready:
+            self._state.active_tree_nodes = []
             return
         self._algorithm.data_array = self._dataset[array_active]
         self._state.coordinates = list(self.data_array.coords.keys())
+        self._state.active_tree_nodes = [array_active]
 
     def bind_x(self, grid_x_array, **kwargs):
         self.algorithm.x = grid_x_array
--- a/pan3d/viewer/engine.py
+++ b/pan3d/viewer/engine.py
@@ -169,8 +169,8 @@ class MeshViewer:
         if not self._state.array_active:
             return
         try:
-            self.mesher.algorithm.Update()
             self.mesher.validate_mesh()
+            self.mesher.algorithm.Update()
             self.actor = self.plotter.add_mesh(
                 self.mesher.algorithm,
                 show_edges=self._state.view_edge_visiblity,
--- a/pan3d/viewer/engine.py
+++ b/pan3d/viewer/engine.py
@@ -74,7 +74,10 @@ class MeshBuilder:
         self._state.error_message = None
 
     def validate_mesh(self):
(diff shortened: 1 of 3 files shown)


pan3d/viewer/ui.py
```diff
@@ -16,204 +16,187 @@ def initialize(server):
...
-                            ctrl.get_render_window(),
-                            # v_show="view_mode === 'view_grid'",
-                            interactive_ratio=1,
-                        ) as vtk_view:
-                            ctrl.view_update = vtk_view.update
-                            ctrl.reset_camera = vtk_view.reset_camera
+                    with vtk.VtkRemoteView(
+                        ctrl.get_render_window(),
+                        interactive_ratio=1,
+                    ) as vtk_view:
```
banesullivan, line 197: You may want to use PyVista's views: https://github.com/pyvista/pyvista/blob/main/pyvista/trame/views.py

- `from pyvista.trame.views import PyVistaRemoteView`
- `from pyvista.trame.ui import plotter_ui`

annehaley: Thanks! Changed in [08db4f5](https://github.com/Kitware/pan3d/pull/23/commits/08db4f5c93448598b8bf9e360f7f2b2ec60e083a)

annehaley: It looks like `plotter_ui` returns a `PyVistaRemoteView` object, so I only imported and directly used `plotter_ui`.

pan3d/viewer/engine.py
```diff
@@ -139,15 +150,30 @@ def __init__(self, server, mesher):
...
-            clim=self.mesher.data_range,
-            **kwargs
-        )
-        self.plotter.view_isometric()
+        if not self._state.array_active:
+            return
+        try:
+            try:
+                # check for valid data
+                self.mesher.algorithm.RequestData(None, None, None)
```
banesullivan, line 158: Never call `RequestData()` directly on any VTK algorithm. Simply call `Update()`

annehaley: Thank you! I had been looking for a function like that! However, it doesn't seem to raise any exceptions for us to show to the user when the update fails. For example, if the user selects one axis, the `RequestData` function was able to give us an error saying more axes are needed for a render. Is there some other function we can call to get such errors to the user's eyes?

pan3d/viewer/engine.py
```diff
@@ -139,15 +150,30 @@ def __init__(self, server, mesher):
...
+            except AttributeError:
+                # TODO: AttributeError will be raised when RequestData
+                # is otherwise successful because we pass None to outputData
+                # We can ignore this and continue, but it would be better to pass
+                # the proper type to outputData
+                pass
+            self.mesher.algorithm.Modified()
+            self.plotter.clear()
```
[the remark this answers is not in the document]
annehaley: The `Modified` and `clear` calls were there before, I just left them in. But I can take them out here.

pan3d/viewer/engine.py
banesullivan: > However, it doesn't seem to raise any exceptions for us to show to the user when the update fails

This is because `RequestData()` is called by the VTK pipeline directly and error handling is managed on the VTK c++ side. This is why I have the `try/except` block that does something with the error message directly.

> For example, if the user selects one axis, the RequestData function was able to give us an error saying more axes are needed for a render.

Perhaps we should do this check outside of the algorithm to make sure the algorithm's inputs are valid at first

> Is there some other function ...

I think these are the errors you are dealing with:

https://github.com/pyvista/pyvista-xarray/blob/2c76baa986f30ea946cdf5cf51628f54643b35ef/pvxarray/rectilinear.py#L49-L57

Perhaps there's a way to implement a similar check here

annehaley: It works if I add these lines to the `validate_mesh` function: [7b56cf0](https://github.com/Kitware/pan3d/pull/23/commits/7b56cf004cdcc38542aa03d9f660ae09011ba37e). Is this good design for fixing the problem?

pan3d/viewer/engine.py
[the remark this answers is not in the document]
annehaley: Removed in [837fdc5](https://github.com/Kitware/pan3d/pull/23/commits/837fdc5bf0c70a6030b8f158fff21c9e60c680e9). It works fine without these lines.

jourdain (approved):
At a high level it seems good. Didn't look at it too closely but didn't see anything problematic.

PR #24, opened 2023-07-12
Incorporate Pangeo Forge data

annehaley:
This PR accomplishes using remote datasets in Pan3D. With this PR, the app fetches a list of available datasets from Pangeo Forge feedstocks to populate the dataset select dropdown. Some example dataset renders are shown below.

This PR rolls back the feature of automatically calling `reset` after every update to the meshing algorithm (axis or resolution inputs). Some data is too big to render and will cause the whole app to hang (as xarray dask tasks do not relinquish control to the event loop). The current discussed solution is to allow the user to decide when to attempt to render. The new reset button says "Apply & Render ([size])" where size is the number of bytes in the current selected data array.

[noaa-coastwatch-geopolar-sst](https://pangeo-forge.org/dashboard/feedstock/3)
image

[WOA-1degree-monthly](https://pangeo-forge.org/dashboard/feedstock/6)
image


johnkit:
I tried using this branch to view the [noaa-coastwatch-geopolar-sst](https://pangeo-forge.org/dashboard/feedstock/3) (after installing fsspect and dask) but I killed the app after a few minutes where it seemed to be stuck busy.

One observation: The viewer did not display a slider for the time axis. Is it possible the viewer is trying to pull all of the time steps? IIRC somewhere during the viewer's development, it downloaded the first time step and displayed a slider that would download time steps on demand. Is there any way to check what the current code is doing?

annehaley:
One avenue for future work would be adding back the feature to automatically call `reset` after changes are made to the meshing algorithm, but this would require some prerequisite size check or the capability to kill a task after it has started. 

With this PR, render tasks are on their own event loop to prevent holding up the Trame app, but within that event loop a task cannot truly be stopped once it has started. If a user asks for a render in state A and that render starts, and then the user immediately asks for a render in state B, the render for state A must complete before the render for state B can begin. The callback for the state A render is cancelled, but the task will not relinquish control to the event loop (and therefore will not stop) until it has completed.

Under this implementation with asyncio and threads, we cannot truly kill a task once it has started. If a subprocess implementation were used, this could become killable, but there are some problems with this that I couldn't find a work-around for. The subprocess cannot be forked because the xarray dask function is not fork-safe, and it cannot be spawned because the pyvista dataset object is not pickleable. 

We currently have no solution for if the user decides to render something that causes the app to hang indefinitely.  One would need to kill the server and restart it. @johnkit decided to ask the xarray folks if this is a known problem with some solution or workaround.

jourdain (approved):

PR #25, opened 2023-07-12
Fix Github actions

annehaley:


jourdain:
@alesgenova

.github/workflows/docker_image.yml
```diff
@@ -3,9 +3,6 @@ on:
   workflow_dispatch:
   push:
     tags: "*"
-    branches:
-      - main
-  pull_request:
```
banesullivan, line 8: I have a personal preference to keep these so that experimental changes can be easily checkout out

src/docker/Dockerfile
```diff
@@ -2,7 +2,8 @@ FROM kitware/trame
 
 RUN apt-get update && \
     apt-get install --no-install-recommends --yes \
-        libosmesa6-dev \
+    libosmesa6-dev \
+    git \
```
jourdain, line 6: I don't think you need git anymore

jourdain (approved):

.github/workflows/docker_image.yml
annehaley: Ok I put them back in [f46dd70](https://github.com/Kitware/pan3d/pull/25/commits/f46dd7065b01f2d3042c5c1264f675fd80060040)

src/docker/Dockerfile
annehaley: Ok, removed in [46943b0](https://github.com/Kitware/pan3d/pull/25/commits/46943b000034de32534e682eeb8d2321665f9737)

banesullivan: Should we use the new vtk-osmesa wheels to avoid libosmesa too?

ref https://discourse.vtk.org/t/status-update-vtk-python-wheels/11212

.github/workflows/docker_image.yml
annehaley: Glad it's turned back on, because now we can see there's actually a failure. It's failing on the following command in the dockerfile:

```
COPY --chown=trame-user:trame-user ./server /deploy/server
```

Do you know what the intent of this line was? The server folder doesn't exist now.

banesullivan: This is because you removed the `build.Dockerfile`. There is a specific build process for Trame. cc @jourdain

annehaley: I had thought it was an overlooked redundancy since the content of the two dockerfiles were very similar. But since it is functional, I've put it back in [2d41f8c](https://github.com/Kitware/pan3d/pull/25/commits/2d41f8c5f9f3db2330d31bfb43e012c0768ccde9). Thanks for explaining.

annehaley: @banesullivan a different error has appeared with that dockerfile put back. It looks like it's attempting to activate a venv that doesn't exist. Does this folder need to be created?

jourdain: So trame allow to build that server directory by running a command line with docker. And another one bundle that generated directory into a docker image.

You can merge the build process inside a single docker image by copying the deploy directory and do a something like that

```
RUN /opt/trame/entrypoint.sh build
```

annehaley:
@jourdain Could you take a look at [377dbb9](https://github.com/Kitware/pan3d/pull/25/commits/377dbb96b6a654d7faae44fa37e487cf92a29032)? Is this what you meant by condensing the build into a single dockerfile? I was trying to follow https://github.com/Kitware/trame/blob/master/examples/deploy/docker/SingleFile/Dockerfile.

jourdain:
yes. Does it work?

annehaley:
> yes. Does it work?

Yes, all tests pass now :+1:

PR #26, opened 2023-07-20
Promote src contents to top-level, delete other modules

annehaley:


alesgenova (approved):
Looks good to me!
We should create a tag for the current `main` branch so we can quickly find the code that is being removed here if we need it as a reference in the future

annehaley:
I have created a branch called `legacy-copy` which has a copy of the current `main`.

jourdain (approved):

PR #28, opened 2023-08-16
Refactor to create Pan3DViewer class

annehaley:
@jourdain and I discovered that a class-based implementation would allow the user to request the gui in a jupyter notebook (which would more reliably output the interface than the previous `show` method). This provided an opportunity to condense MeshViewer and MeshBuilder into one class and consolidate the application logic. 

This format also made it far easier to specify and load an initial state for the app, so I got a head start on the bookmarking feature on this branch.

Lastly, I needed to confirm these functions worked in jupyter notebooks as well, so this branch also includes new examples in the examples folder.


pan3d/viewer.py
```diff
@@ -0,0 +1,265 @@
+import json
+import os
+import pyvista
+from pvxarray.vtk_source import PyVistaXarraySource
+import xarray
+
+from trame.decorators import TrameApp, change
+from trame.ui.vuetify import SinglePageWithDrawerLayout  # TODO: upgrade to vuetify 3
```
jourdain, line 8: The PyVista gui from trame are based on vue2 (They use vuetify2 for their buttons) so you will run into issue.

pan3d/viewer.py
```diff
@@ -0,0 +1,265 @@
...
+        if dataset_path:
+            self.set_dataset_path(dataset_path=dataset_path)
+        if state:
+            self.state.update(state)
+
+        # Fix version of vue
+        server.client_type = "vue2"  # TODO: upgrade to vue3
+        # Build GUI
+        self.layout = SinglePageWithDrawerLayout(self.server)
+        viewer_ui(self.layout, self.ctrl)
```
jourdain, line 46: I don't think the UI needs to be externalized. Also the layout has the server, so you have access to the controller from there as well. But if you have everything defined within the class, you don't need the controller. You can directly by your self.method.

pan3d/viewer.py
annehaley: Yes, I tested this and did get a working version with a quick fix in the pyvista code. I filed an issue for the pyvista developers here: https://github.com/pyvista/pyvista/issues/4776. I have a vuetify 3 version of the ui that would work if we can help the pyvista folks address this (hopefully in a more robust manner than my quick fix). But for now I'll leave a todo.

pan3d/ui.py
```diff
@@ -1,16 +1,9 @@
...
 
-# Create single page layout type
-# (FullScreenPage, SinglePage, SinglePageWithDrawer)
-def initialize(server):
-    state, ctrl = server.state, server.controller
-    state.trame__title = "Pan3D Viewer"
-
-    with SinglePageWithDrawerLayout(server) as layout:
-        # Toolbar
+def initialize(layout, ctrl):
```
jourdain, line 5: That could be part of the main class

annehaley: I agree but I also think the file would be pretty long if we combine them. Is it acceptable to keep them separate and just restructure what we send to the ui creation function?

annehaley:
@jourdain is there a known reason why the Windows tests are consistently getting segmentation faults?

jourdain:
No idea why Windows is failing

annehaley:
I found out why the segmentation fault was occuring. I should not have put `server.start()` in the pytest.

examples/jupyter/pangeo_forge.ipynb
```diff
@@ -0,0 +1,73 @@
...
+    "    't_index': 5,\n",
+    "    'resolution': 0.1,\n",
+    "    'main_drawer': False,\n",
+    "}\n",
+    "builder = DatasetBuilder(dataset_path=dataset_path, state=state)\n",
+    "\n",
+    "# Show viewer in cell output\n",
+    "await builder.layout.ready\n",
+    "builder.viewer"
+   ]
```
jourdain, line 36: This is weird

```python
await builder.layout.ready
builder.viewer
```

I would assume the viewer is the layout so I would assume

```python
await builder.viewer.ready
builder.viewer
```

pan3d/__main__.py
```diff
@@ -0,0 +1,17 @@
...
+    description="Launch the Pan3D Dataset Viewer / Builder",
+)
+
+parser.add_argument("-b", "--config_path")
+parser.add_argument("-d", "--dataset_path")
+
+args = parser.parse_args()
+
+viewer = DatasetBuilder(dataset_path=args.dataset_path)
+if args.config_path:
```
jourdain, line 15: builder...

examples/jupyter/pangeo_forge.ipynb
annehaley: Do you mean that the await statement can be removed?

pan3d/dataset_builder.py
```diff
@@ -0,0 +1,284 @@
...
+    def state(self):
+        return self.server.state
+
+    @property
+    def ctrl(self):
+        return self.server.controller
+
+    @property
+    def viewer(self):
+        return self.layout
```
jourdain, line 81: This could encapsulate line 50-65 with some lazy loading logic.

pan3d/__main__.py
annehaley: Ok, changed in [478af46](https://github.com/Kitware/pan3d/pull/28/commits/478af46b55a1d8e87be39ade88fadfcab4cee527)

pan3d/dataset_builder.py
jourdain:
```python

@property
    def viewer(self):
        if self.layout is None:
              # l50-65
        return self.layout
```

pan3d/ui/axis_selection.py
```diff
@@ -0,0 +1,111 @@
+from trame.widgets import html, vuetify
+
+
+class AxisSelection(vuetify.VNavigationDrawer):
+    def __init__(self):
+        super().__init__(
+            value=("array_active",),
```
jourdain, line 7: Should __"array_active"__ be an actual parameter rather than fixed like that?

annehaley: Do you mean using `v_model`? We wouldn't want `array_active` to be set to false when the drawer is collapsed.

pan3d/dataset_builder.py
annehaley: Ok, changed in [a13307f](https://github.com/Kitware/pan3d/pull/28/commits/a13307f4fb78d9a4a6d0b29dda9d3cac963df7f1)

examples/jupyter/pangeo_forge.ipynb
jourdain: I mean why do you have 2 things that are the same but with different name. It gets confusing. Stick with a single name.

annehaley: Oh I see now, they looked similar enough that I missed the diff

pan3d/dataset_builder.py
```diff
@@ -0,0 +1,284 @@
...
+                with self.layout.content:
+                    plot_view = plotter_ui(
+                        self.ctrl.get_plotter(),
+                        interactive_ratio=1,
+                    )
+                    self.ctrl.view_update = plot_view.update
+                    self.ctrl.reset_camera = plot_view.reset_camera
+                    RenderArea(plot_view=plot_view)
+                AxisSelection()
+        return self.layout
```
jourdain, line 81: Maybe we should make layout internal by using `self._layout`

pan3d/ui/axis_selection.py
jourdain: What I'm saying is that the macro component should not make the choice of variable names. 

```python
class AxisSelection(vuetify.VNavigationDrawer):
    def __init__(self, drawer_visible):
        super().__init__(
            value=(drawer_visible,),
```

jourdain: But if you want to keep those fixed on purpose that is fine. You are just preventing yourself of really leveraging the flexibility of what those "macro" components means. 

So far, you are using them like static function rather than reusable piece of UI.

examples/jupyter/pangeo_forge.ipynb
annehaley: Changed in [d65129e](https://github.com/Kitware/pan3d/pull/28/commits/d65129e719ede4e0114ca7b3be73100391ff2017)

pan3d/ui/main_drawer.py
```diff
@@ -0,0 +1,71 @@
+from trame.widgets import html, vuetify
+
+
+class MainDrawer(html.Div):
+    def __init__(self):
```
jourdain, line 5: Looks like fully static code. No parametrization at all.

pan3d/ui/axis_selection.py
annehaley: Should that apply to all state variables, like the axis vars, etc?

pan3d/ui/render_area.py
```diff
@@ -0,0 +1,15 @@
+from trame.widgets import html, vuetify
+
+
+class RenderArea(html.Div):
+    def __init__(self, plot_view=None):
```
jourdain, line 5: The plot_view won't work

pan3d/ui/render_area.py
```diff
@@ -0,0 +1,15 @@
...
+        super().__init__(
+            v_show="array_active",
+            style="height: 100%; position: relative; width: calc(100% - 300px)",
+        )
+        with self:
+            vuetify.VBanner(
+                "{{ error_message }}",
+                v_show=("error_message",),
+            )
+            plot_view
```
jourdain, line 15: As far as I know like 15 is a NoOp. Are you expecting something from that?

jourdain (changes requested):

pan3d/ui/render_area.py
annehaley: I was hoping I could do that because I had to move the plotter creation back into the main class since it requires the controller. Since passing it through doesn't work, I'll just dissolve this class back into the main one. This class can't have any reason to exist if it's not doing the job of bringing the plotter ui into the layout.

pan3d/ui/main_drawer.py
annehaley: Yes, while it isn't the typical use case for a component class, I would still want to avoid dissolving this into the main class. I am kind of treating it as function-based code simply to avoid a very large file and to have some conceptual separation of this code. 
If passing in state var names helps to legitimize this separation, I can do that. The expectation is that these var names will never have any other value, so I could put the expected string as defaults, so the API where I've used these "classes" wouldn't have to change either. But it would still technically be possible to reuse the class with different state var names. Is this an acceptable approach?

pan3d/dataset_builder.py
annehaley: Ok changed in [75c2ec7](https://github.com/Kitware/pan3d/pull/28/commits/75c2ec71f08e87d7a576740f3bc55b067d1904cb)

pan3d/ui/render_area.py
annehaley: Changed in [0c36b45](https://github.com/Kitware/pan3d/pull/28/commits/0c36b45c413e22cc2a0963dde8eb4f875d848413)

pan3d/ui/render_area.py
jourdain: Well you can do it but not like that.

jourdain: See 

```python
                    
with RenderArea() as view:
    with view.content:
        plot_view = plotter_ui(
            self.ctrl.get_plotter(),
            interactive_ratio=1,
        )
        self.ctrl.view_update = plot_view.update
        self.ctrl.reset_camera = plot_view.reset_camera


```

jourdain: Think of what you do in vue.js with slots... This is the same concept.

annehaley: I see. Do you think it's still worth separating a component for it?

jourdain: If it is tiny, probably not... But that split is up to you to decide what make sense.

Here is an example of one that I just made

```python
class VCard(vuetify.VCard):
    def __init__(self, show_name, title=None, **kwargs):
        super().__init__(**kwargs)
        with self:
            self.header = vuetify.VCardTitle(title, classes="py-2")
            with self.header:
                vuetify.VSpacer()
                vuetify.VIcon(
                    v_text=f"{show_name} ? 'mdi-chevron-up':'mdi-chevron-down'",
                    click=f"{show_name} = !{show_name}",
                )
            vuetify.VDivider(v_show=show_name)
            self.content = vuetify.VCardText(v_show=(show_name, True))

```

pan3d/ui/main_drawer.py
annehaley: @jourdain Does the approach in [5b7331b](https://github.com/Kitware/pan3d/pull/28/commits/5b7331b4dadd55dcaaa3d2650ee192bb603740af) make this better?

jourdain: yes, perfect!

pan3d/ui/axis_selection.py
annehaley: Changed in [5b7331b](https://github.com/Kitware/pan3d/pull/28/commits/5b7331b4dadd55dcaaa3d2650ee192bb603740af)

pan3d/ui.py
annehaley: Resolving because the UI components don't use the controller anymore

pan3d/ui/render_area.py
annehaley: I'll leave it in the main component for now, since we moved the UI creation to the lazily-loaded viewer function. It's pretty tiny as just the display of the error and the plotter_ui object from pyvista, so it doesn't warrant its own file. Having the other three components separated makes the viewer code a reasonable size.

jourdain: That was the intent so we could have a grasp of the big picture of what the viewer looks like (drawer, content, ...)

pan3d/__main__.py
```diff
@@ -0,0 +1,18 @@
...
+parser.add_argument("-b", "--config_path")
+parser.add_argument("-d", "--dataset_path")
+
+args = parser.parse_args()
+
+builder = DatasetBuilder(dataset_path=args.dataset_path)
+if args.config_path:
+    builder.import_config(args.config_path)
+viewer = builder.viewer
+builder.server.start()
```
jourdain, line 18: I would replace 

```python
viewer = builder.viewer
builder.server.start()
```

by 

```python
builder.viewer.server.start()
```

jourdain: In some way, I would make builder not really "trame server" dependent. But we can work on that piece later on especially since it will be trivial.

jourdain: All the refactoring your are doing now, will make sure a change a piece of cake down the road.

jourdain (approved):

pan3d/__main__.py
annehaley: Good idea, changed in [0a844e2](https://github.com/Kitware/pan3d/pull/28/commits/0a844e2c9791ffe7cf23a5b37e165d91ba1bf8b4)

annehaley:
@jourdain Sorry could you leave your review again? Making the extra commit dismissed the previous approval.

jourdain (approved):

PR #29, opened 2023-08-25
Axis selector redesign

annehaley:
## Summary
This PR creates a new UI for assigning coordinates to the available rendering axes.
image

## Details
The new UI is composed of two new components, `AxisSelection` and `CoordinateConfigure`. The axis selection component is the organizational wrapper that shows the placement of each coordinate configure component. The coordinate configure component is a collapsible panel that allows the user to view the attributes of the coordinate, specify slicing, and assign the coordinate to an axis. 

The display of a coordinate configure component changes slightly when a coordinate is assigned to the T axis. Rather than seeing a slicing section (with inputs for start, stop, and step), the user sees a slider to select a single index.
image

## Results

https://github.com/Kitware/pan3d/assets/44912689/d889ba1a-c637-4d14-bc0c-5e7dd45d9f88


pan3d/__main__.py
```diff
@@ -1,17 +1,18 @@
...
 
-parser = ArgumentParser(
+parser = argparse.ArgumentParser(
     prog="Pan3D",
     description="Launch the Pan3D Dataset Builder",
 )
 
 parser.add_argument("-b", "--config_path")
 parser.add_argument("-d", "--dataset_path")
+parser.add_argument("-s", "--server", action=argparse.BooleanOptionalAction)
```
jourdain, line 11: Does trame already take care of that?

jourdain: That is because you don't use the CLI from trame... But your approach is probably better if we really make trame optional for the viewer section down the road...

pan3d/dataset_builder.py
```diff
@@ -116,6 +141,37 @@ def data_range(self):
...
+    # -----------------------------------------------------
+    # UI bound methods
+    # -----------------------------------------------------
+
+    def coordinate_select_axis(self, coordinate_name, current_axis, new_axis, **kwargs):
+        if getattr(self.state, current_axis):
+            setattr(self.state, current_axis, None)
+        if new_axis and new_axis != "undefined":
+            setattr(self.state, new_axis, coordinate_name)
+        self.mesh_changed()
```
jourdain, line 153: Can you explain what you are trying to do here? That seems convoluted.

jourdain: https://trame.readthedocs.io/en/latest/core.state.html

pan3d/dataset_builder.py
```diff
@@ -171,7 +231,34 @@ def set_dataset_path(self, dataset_path, **kwargs):
...
+                    "attrs": coord_attrs,
+                    "size": da.coords[key].size,
+                    "range": [array_min, array_max],
+                    "start": array_min,
+                    "stop": array_max,
+                    "step": 1,
+                }
+            )
+            self.state.expanded_coordinates.append(key),
+        self.auto_select_coordinates()
```
jourdain, line 261: Since you are editing in place `expanded_coordinates` and `coordinates`. Don't you need to mark them dirty?

pan3d/dataset_builder.py
```diff
@@ -225,6 +307,23 @@ def on_set_scale(self, x_scale=None, y_scale=None, z_scale=None, **kwargs):
...
+                # If accepted name is longer than one letter, look for contains match
+                name_match = [
+                    coordinate["name"]
+                    for accepted in accepted_names
+                    if (len(accepted) == 1 and accepted == name)
+                    or (len(accepted) > 1 and accepted in name)
+                ]
+                if len(name_match) > 0:
+                    state_update[axis] = name_match[0]
+        if len(state_update.keys()) > 0:
```
jourdain, line 324: In python I think you can just do `len(state_update)`

jourdain (approved):
Small code improvements could be made. But those fixes could be done in a separate PR.

pan3d/dataset_builder.py
annehaley: When a user selects an option in the "Assign Axis" dropdown for any coordinate, the previously assigned axis should be set to None (if there was a previous assignment) and the newly assigned axis should be set to the coordinate name. 

Is your suggestion to use `self.state.update(...)` instead of `setattr(self.state, ...)`?

pan3d/dataset_builder.py
annehaley: I hadn't noticed a case in which they don't update, but I'll add that line in anyway to protect against it.

jourdain: Probably something updates the content or marks them dirty somewhere else. But it is good practice to be explicit if you aim that variable to be properly shared with the client.

pan3d/dataset_builder.py
annehaley: Changed in [b670a16](https://github.com/Kitware/pan3d/pull/29/commits/b670a16a70e27b9de13458af5130bb75ee8c1f0f)

pan3d/dataset_builder.py
jourdain: I would think something more like `self.state.new_axis = coordinate_name`

pan3d/dataset_builder.py
annehaley: Added in [910fe0c](https://github.com/Kitware/pan3d/pull/29/commits/910fe0c13c7e97599837f028138bcb959a66f4dc)

pan3d/dataset_builder.py
jourdain: unless I'm missing something

jourdain: Basically, I'm not sure why you use setattr and getattr...

annehaley: `new_axis` is a variable which will contain the name of a state attribute, it is not a state attribute itself. Its value would be one of `x_array`, `y_array`, `z_array`, or `t_array`.

annehaley: (the same applies to `current_axis`)

jourdain: then `self.state[new_axis] = coordinate_name`

jourdain: The first line of the State class API documentation

> Flexible dictionary managing a server shared state. Variables can be accessed with either the [] or . notation.

jourdain: And the usage of the state is also covered here https://kitware.github.io/trame/docs/cheatsheet.html#State

annehaley: Ok, I'll use bracket notation then, if that is preferred: changed in [5304ceb](https://github.com/Kitware/pan3d/pull/29/commits/5304ceb985f12378e6a2afde69136a19ba95a165)

jourdain (approved):

PR #30, opened 2023-09-01
Vue 3 migration

annehaley:
Before going too far on the Axis Selector redesign (#29) in Vue 2, I'd like to migrate to Vue 3 and use that API going forward. My pyvista PR was just merged (https://github.com/pyvista/pyvista/pull/4811), and a release will be going out soon.

This PR switches to the Vue 3 client type, and once the pyvista release is made, the pyvista `plotter_ui` function we use will return a compatible interface. 

** Requires pyvista >= 0.42.0 **

### DIFF
--- a/pan3d/dataset_builder.py
+++ b/pan3d/dataset_builder.py
@@ -6,9 +6,10 @@ from pvxarray.vtk_source import PyVistaXarraySource
 from pyvista.trame.ui import plotter_ui
 
 from trame.decorators import TrameApp, change
-from trame.ui.vuetify import SinglePageWithDrawerLayout  # TODO: upgrade to vuetify 3
+from trame.ui.vuetify3 import SinglePageWithDrawerLayout
 from trame.app import get_server
-from trame.widgets import html, vuetify
+from trame.widgets import html, client
+from trame.widgets import vuetify3 as vuetify
 
 from pan3d.ui import AxisSelection, MainDrawer, Toolbar
 from pan3d.utils import initial_state, run_singleton_task
@@ -24,7 +25,7 @@ class DatasetBuilder:
             server = get_server(server)
 
         # Fix version of vue
-        server.client_type = "vue2"  # TODO: upgrade to vue3
+        server.client_type = "vue3"
         self.server = server
         self._layout = None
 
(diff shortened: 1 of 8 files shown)


pan3d/dataset_builder.py
```diff
@@ -63,18 +66,18 @@ def viewer(self):
             # Build UI
             self._layout = SinglePageWithDrawerLayout(self.server)
             with self._layout as layout:
+                client.Style(open(f"{DIR}/ui/custom.css").read())
```
jourdain, line 69: You should use pathlib and Path

```python
from pathlib import Path

BASE_DIR = Path(__file__).parent
CSS_FILE = BASE_DIR / "ui" / "custom.css"

client.Style(CSS_FILE.read_text())
```

pan3d/dataset_builder.py
```diff
@@ -172,6 +175,11 @@ def set_dataset_path(self, dataset_path, **kwargs):
         if len(self.state.data_attrs) > 0:
             self.state.show_data_attrs = True
         self.state.coordinates = []
+        self.state.x_array = None
+        self.state.y_array = None
+        self.state.z_array = None
+        self.state.t_array = None
+        self.state.t_index = 0
```
jourdain, line 182: Can those state var be initialized by the UI? Or the pieces that need/expect them?

jourdain (changes requested):
I'm guessing the vuetify2/3 api change is fully working.

pan3d/dataset_builder.py
annehaley: They could be initialized by the UI, but we also want to clear their values every time the selected dataset is changed. Having these statements here is both an initialization and a reset. We could copy these statements to initialize them in the UI, but it might be redundant since we need these here too.

jourdain: ok sound good.

Maybe you can condense it like

`self.state.update(dict(y_array=None, z_array=None, t_array=None, t_index=0))`

pan3d/dataset_builder.py
annehaley: Thanks, changed in [81ab442](https://github.com/Kitware/pan3d/pull/30/commits/81ab442f8daba987d66b386289633fd0e4a6699c)

pan3d/dataset_builder.py
annehaley: Ok, changed in [1cb2197](https://github.com/Kitware/pan3d/pull/30/commits/1cb21978c4018dc1f92cc2e468f5ab332190ce7a)

jourdain (approved):

PR #31, opened 2023-09-05
Fix `pan3d-viewer` console script

annehaley:
This update should have been made in #28 because the old script location no longer exists.


jourdain (approved):
The relationship between the DataSetBuilder and the viewer is still entangled. We should revisit that at some point, but for now. That is good to go.

PR #32, opened 2023-09-07
Allow other xarray engines for other file types

annehaley:
I was able to test out the `pflotran` dataset from @johnkit, and I can confirm it works with Pan3D. We just need a few lines to support the `.nc` version of the data as well. 

image

### DIFF
--- a/pan3d/dataset_builder.py
+++ b/pan3d/dataset_builder.py
@@ -133,10 +133,14 @@ class DatasetBuilder:
             ):
                 self.state.more_info_link = available_dataset["more_info"]
         if "https://" in dataset_path or os.path.exists(dataset_path):
-            # Assumes zarr store
+            engine = None
+            if ".zarr" in dataset_path:
+                engine = "zarr"
+            if ".nc" in dataset_path:
+                engine = "netcdf4"
             try:
                 self.dataset = xarray.open_dataset(
-                    dataset_path, engine="zarr", chunks={}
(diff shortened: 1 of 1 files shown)


pan3d/dataset_builder.py
```diff
@@ -133,10 +133,14 @@ def set_dataset_path(self, dataset_path, **kwargs):
...
-            # Assumes zarr store
+            engine = None
+            if ".zarr" in dataset_path:
+                engine = "zarr"
+            if ".nc" in dataset_path:
+                engine = "netcdf4"
             try:
                 self.dataset = xarray.open_dataset(
-                    dataset_path, engine="zarr", chunks={}
+                    dataset_path, engine=engine, chunks={}
```
jourdain, line 143: Is "engine" a good name for it? I'm not necessary asking to change but just wondering between

- engine
- format
- data_format
- data_type
- dataset_format

annehaley: All of these are good, and I'm not opposed to changing it. The reason I called it engine is because the value must be one of the options for engine in [xarray.oppen_dataset](https://docs.xarray.dev/en/stable/generated/xarray.open_dataset.html). I'll make it your call though if you think another one works best.

jourdain: Since it is a direct mapping, we should keep it as **engine**

jourdain (approved):

PR #33, opened 2023-09-20
Add Import and Export buttons

annehaley:
This PR adds the import and export features to Pan3D. The video attached below captures the new behavior.
https://github.com/Kitware/pan3d/assets/44912689/fd4190be-26e1-4ec8-abb3-bb0c951371b7

In this example, two config files are used. First, an import is performed on the second config file in the `examples` folder. The contents of that file are as follows:

```
{
    "data_origin": "eraint_uvz",
    "data_array": {
        "active": "z",
        "x": "longitude",
        "y": "latitude",
        "z": "level",
        "t": "month"
    },
    "data_slices": {
        "latitude": [
            -90,
            0,
            10
        ],
        "longitude": [
            -180,
            0,
            10
        ],
        "level": [
            200,
            850,
            1
        ],
        "month": [
            0,
            7,
            1
        ]
    },
    "ui": {
        "expanded_coordinates": [
            "longitude",
            "latitude",
            "level",
            "month"
        ],
        "main_drawer": true
    }
}
```

Then, some modifications are made to the slices and an export is performed, saved as `pan3d_state.json`. The contents of the exported file are as follows:

```
{
    "data_origin": "eraint_uvz",
    "data_array": {
        "active": "z",
        "x": "longitude",
        "y": "latitude",
        "z": "level",
        "t": "month",
        "t_index": "1"
    },
    "data_slices": {
        "longitude": [
            -180,
            90,
            2
        ],
        "latitude": [
            -90,
            90,
            2
        ],
        "level": [
            200,
            850,
            1
        ],
        "month": [
            1,
            7,
            1
        ]
    },
    "ui": {
        "main_drawer": true,
        "expanded_coordinates": []
    }
}
```

Finally, the video shows switching to a different dataset before performing an import of the newly exported `pan3d_state.json`.


pan3d/dataset_builder.py
```diff
@@ -386,12 +411,81 @@ def mesh_updated(exception=None):
...
+
+        if slices_config:
+            coordinates = self.state.coordinates
+            for coordinate in coordinates:
+                if coordinate["name"] in slices_config:
+                    start, stop, step = slices_config[coordinate["name"]]
+                    coordinate["start"] = start
+                    coordinate["stop"] = stop
+                    coordinate["step"] = step
+            self.state.coordinates = coordinates
```
jourdain, line 446: remove `self.state.coordinates = coordinates`

pan3d/dataset_builder.py
```diff
@@ -386,12 +411,81 @@ def mesh_updated(exception=None):
...
+            self.state.dirty("coordinates")
+
+        if ui_config:
+            self.state.update(ui_config)
+            self.state.dirty(*ui_config.keys())
```
jourdain, line 451: Do you need to mark the ui props dirty? I'm not sure why.

jourdain: Were you going to prefix ui state vars by `ui_xxxx`?

pan3d/dataset_builder.py
```diff
@@ -386,12 +411,81 @@ def mesh_updated(exception=None):
...
-        pass
+        self.state.update({"dialog_shown": None, "selected_config_file": None})
+
+    def export_config(self, config_file=None):
+        config = {}
+        config["data_origin"] = self.state.dataset_path
+        config["data_array"] = {"active": self.state.active_array}
+
+        for axis in ["x", "y", "z", "t"]:
+            if self.state[axis + "_array"]:
```
jourdain, line 461: Doing string operations with a + is not great. If would be better to use `f"{axis}_array"`

jourdain: Also, it would be better to use a prefix rather than a postfix.

pan3d/dataset_builder.py
```diff
@@ -386,12 +411,81 @@ def mesh_updated(exception=None):
...
+                    coordinate.get("start", 0),
+                    coordinate.get("stop", -1),
+                    coordinate.get("step", 1),
+                ]
+                config["data_slices"][coordinate["name"]] = coordinate_slice
+
+        for state_var in ["main_drawer", "expanded_coordinates"]:
+            if "ui" not in config:
+                config["ui"] = {}
+            config["ui"][state_var] = self.state[state_var]
```
jourdain, line 485: `config["ui"][state_var] = self.state[f"ui_{state_var}"]`

pan3d/dataset_builder.py
```diff
@@ -386,12 +411,81 @@ def mesh_updated(exception=None):
...
+
+        if config_file:
+            with open(config_file, "w") as f:
```
jourdain, line 488: you can use Path from pathlib to write the file in 1 line.

pan3d/ui/coordinate_configure.py
```diff
@@ -42,7 +47,7 @@ def __init__(
                                     "%s?.size - 1" % coordinate_info,
                                     0,
                                 ),
-                                step=1,
+                                step=("%s?.step" % coordinate_info, 0),
```
jourdain, line 50: f string?

pan3d/ui/coordinate_configure.py
```diff
@@ -55,62 +60,95 @@ def __init__(
                                             "%s?.size - 1" % coordinate_info,
                                             0,
                                         ),
+                                        step=("%s?.step" % coordinate_info, 0),
```
jourdain, line 63: f-string

pan3d/ui/coordinate_configure.py
```diff
@@ -55,62 +60,95 @@ def __init__(
...
+                        ):
+                            vuetify.VTextField(
+                                model_value=("%s?.start" % coordinate_info, 0),
+                                label="Start",
+                                hide_details=True,
+                                density="compact",
+                                type="number",
+                                min=(
+                                    "parseFloat(+ %s?.range[0].toFixed(2))"
+                                    % coordinate_info,
```
jourdain, line 83: f-string

pan3d/ui/coordinate_configure.py
```diff
@@ -55,62 +60,95 @@ def __init__(
...
+                                    0,
+                                ),
+                                max=(
+                                    "parseFloat(+ %s?.stop.toFixed(2))"
+                                    % coordinate_info,
```
jourdain, line 88: f-string

pan3d/ui/coordinate_configure.py
```diff
@@ -55,62 +60,95 @@ def __init__(
...
+                                style="flex-grow: 1",
+                            )
+                            vuetify.VTextField(
+                                model_value=("%s?.stop" % coordinate_info, 0),
+                                label="Stop",
+                                hide_details=True,
+                                density="compact",
+                                type="number",
+                                min=(
+                                    "parseFloat(+ %s?.start.toFixed(2))"
```
jourdain, line 111: f-string

pan3d/ui/coordinate_configure.py
```diff
@@ -55,62 +60,95 @@ def __init__(
...
+                                    % coordinate_info,
+                                    0,
+                                ),
+                                max=(
+                                    "parseFloat(+ %s?.range[1].toFixed(2))"
```
jourdain, line 116: f-string

pan3d/ui/coordinate_configure.py
```diff
@@ -55,62 +60,95 @@ def __init__(
...
+                                    % coordinate_info,
+                                    0,
+                                ),
+                                step="0.01",
+                                __properties=[
+                                    ("min", "min"),
+                                    ("max", "max"),
+                                    ("step", "step"),
+                                ],
```
jourdain, line 125: since you have a 1-to-1 mapping between Python and JS, you should just do `__properties=["min", "max", "step"],`

pan3d/ui/coordinate_configure.py
```diff
@@ -55,62 +60,95 @@ def __init__(
...
+                                input=(
+                                    coordinate_change_slice_function,
+                                    f"[{coordinate_info}.name, 'stop', $event.target.value]",
+                                ),
+                                __events=[("input", "input.prevent")],
+                                style="flex-grow: 1",
+                            )
+                            vuetify.VTextField(
+                                model_value=("%s?.step" % coordinate_info, 0),
```
jourdain, line 134: f-string

pan3d/ui/coordinate_configure.py
```diff
@@ -55,62 +60,95 @@ def __init__(
...
+                                label="Step",
+                                hide_details=True,
+                                density="compact",
+                                type="number",
+                                min="1",
+                                max=("%s?.size" % coordinate_info, 0),
```
jourdain, line 140: f-string

pan3d/ui/coordinate_configure.py
```diff
@@ -55,62 +60,95 @@ def __init__(
...
+                                __properties=[
+                                    ("min", "min"),
+                                    ("max", "max"),
```
jourdain, line 143: just `["min", "max"]`

pan3d/ui/file_select.py
```diff
@@ -0,0 +1,102 @@
...
+        with self:
+            with vuetify.VCardText(v_show=dialog_shown):
+                vuetify.VBtn(
+                    flat=True,
+                    icon="mdi-close",
+                    style="float: right",
+                    click="%s = undefined" % dialog_shown,
+                )
+                vuetify.VCardTitle(
+                    "{{ %s }}" % dialog_message,
```
jourdain, line 33: f-string

pan3d/ui/file_select.py
```diff
@@ -0,0 +1,102 @@
...
+                    v_show=dialog_message,
+                )
+                with html.Div(v_show="!%s" % dialog_message):
+                    vuetify.VCardTitle("{{ %s }} File Select" % dialog_shown)
```
jourdain, line 37: f-string

pan3d/ui/file_select.py
```diff
@@ -0,0 +1,102 @@
...
+                    flat=True,
+                    icon="mdi-close",
+                    style="float: right",
+                    click="%s = undefined" % dialog_shown,
+                )
+                vuetify.VCardTitle(
+                    "{{ %s }}" % dialog_message,
+                    v_show=dialog_message,
+                )
+                with html.Div(v_show="!%s" % dialog_message):
```
jourdain, line 36: f-string

pan3d/ui/file_select.py
```diff
@@ -0,0 +1,102 @@
...
+                                            %s = 'Export complete.'
+                                        })
+                                    });
+                                } catch {
+                                    %s = false;
+                                    window.alert('Your browser does not support selecting a download location. Your download will be made to the default location, saved as pan3d_state.json.')
+                                }
+                            }
+                        """
+                        % (state_export, dialog_message, selected_config_file),
```
jourdain, line 83: f-string

pan3d/ui/file_select.py
```diff
@@ -0,0 +1,102 @@
...
+                            var a = window.document.createElement('a');
+                            a.href = 'data:application/json;charset=utf-8,'  + encodeURIComponent(content);
+                            a.download = 'pan3d_state.json';
+                            a.style.display = 'none';
+                            window.document.body.appendChild(a);
+                            a.click()
+                            window.document.body.removeChild(a);
+                            %s = 'Export complete.'
+                        """
+                        % (state_export, dialog_message),
```
jourdain, line 100: f-string

pan3d/ui/toolbar.py
```diff
@@ -22,17 +25,25 @@ def __init__(
...
                 html.Span("Apply & Render")
                 html.Span("({{ %s }})" % da_size, v_show=da_size)
-            vuetify.VCheckbox(
-                v_model=(view_edge_visibility, True),
-                v_show=array_active,
-                density="compact",
-                true_icon="mdi-border-all",
-                false_icon="mdi-border-outside",
+            vuetify.VBtn(
+                click="%s = 'Export'" % dialog_shown,
```
jourdain, line 35: f-string

pan3d/ui/toolbar.py
```diff
@@ -22,17 +25,25 @@ def __init__(
...
             )
             with vuetify.VBtn(
-                click=reset,
+                click=reset_function,
                 v_show=unapplied_changes,
-                # small=True,
                 variant="tonal",
             ):
                 html.Span("Apply & Render")
                 html.Span("({{ %s }})" % da_size, v_show=da_size)
```
jourdain, line 33: f-string or maybe not here since it will add more `{{}}` which can confuse the code. Then use the explicit format method.

pan3d/ui/toolbar.py
```diff
@@ -22,17 +25,25 @@ def __init__(
...
-                density="compact",
-                true_icon="mdi-border-all",
-                false_icon="mdi-border-outside",
+            vuetify.VBtn(
+                click="%s = 'Export'" % dialog_shown,
+                variant="tonal",
+                text="Export",
+            )
+            vuetify.VBtn(
+                click="%s = 'Import'" % dialog_shown,
```
jourdain, line 40: f-string

pan3d/utils.py
```diff
@@ -65,24 +65,24 @@ async def coroutine():
...
     "t_max": 0,
-    "view_edge_visibility": True,
     "error_message": None,
     "da_size": None,
     "mesh_timeout": 30,
+    "dialog_shown": None,
+    "dialog_message": None,
+    "selected_config_file": None,
+    "state_export": None,
 }
```
jourdain, line 86: All those state variable names should be group and easy to spot which part they belong to (ui_*, da_*, ...)

pan3d/dataset_builder.py
jourdain: assuming we go with the ui_ prefix logic.

jourdain (changes requested):
Some code cleanup for code consistency. 

Where f-string are not convenient, you can use the `"{{ {a} }}".format(a=4)` approached.

pan3d/ui/toolbar.py
annehaley: Changed in [28561cb](https://github.com/Kitware/pan3d/pull/33/commits/28561cb331878bed9f5131df5a01d8307360c41e)

pan3d/ui/toolbar.py
annehaley: Changed in [28561cb](https://github.com/Kitware/pan3d/pull/33/commits/28561cb331878bed9f5131df5a01d8307360c41e)

pan3d/ui/toolbar.py
annehaley: Changed in [28561cb](https://github.com/Kitware/pan3d/pull/33/commits/28561cb331878bed9f5131df5a01d8307360c41e)

pan3d/ui/file_select.py
annehaley: Changed in [28561cb](https://github.com/Kitware/pan3d/pull/33/commits/28561cb331878bed9f5131df5a01d8307360c41e)

pan3d/ui/file_select.py
annehaley: Changed in [28561cb](https://github.com/Kitware/pan3d/pull/33/commits/28561cb331878bed9f5131df5a01d8307360c41e)

pan3d/ui/file_select.py
annehaley: Changed in [28561cb](https://github.com/Kitware/pan3d/pull/33/commits/28561cb331878bed9f5131df5a01d8307360c41e)

pan3d/ui/file_select.py
annehaley: Changed in [28561cb](https://github.com/Kitware/pan3d/pull/33/commits/28561cb331878bed9f5131df5a01d8307360c41e)

pan3d/ui/coordinate_configure.py
annehaley: Changed in [28561cb](https://github.com/Kitware/pan3d/pull/33/commits/28561cb331878bed9f5131df5a01d8307360c41e)

pan3d/ui/coordinate_configure.py
annehaley: Changed in [28561cb](https://github.com/Kitware/pan3d/pull/33/commits/28561cb331878bed9f5131df5a01d8307360c41e)

pan3d/ui/coordinate_configure.py
annehaley: Changed in [28561cb](https://github.com/Kitware/pan3d/pull/33/commits/28561cb331878bed9f5131df5a01d8307360c41e)

pan3d/ui/coordinate_configure.py
annehaley: Changed in [28561cb](https://github.com/Kitware/pan3d/pull/33/commits/28561cb331878bed9f5131df5a01d8307360c41e)

pan3d/ui/coordinate_configure.py
annehaley: Changed in [28561cb](https://github.com/Kitware/pan3d/pull/33/commits/28561cb331878bed9f5131df5a01d8307360c41e)

pan3d/utils.py
annehaley: Variables renamed in [6f497a9](https://github.com/Kitware/pan3d/pull/33/commits/6f497a9be4f3e7e817c6b87aa61a51dcd2ea71bf)

pan3d/ui/coordinate_configure.py
annehaley: Changed in [6088389](https://github.com/Kitware/pan3d/pull/33/commits/608838948dfe0dc7ae087a8bad66d8cdfe9a7aa9)

pan3d/dataset_builder.py
annehaley: Changed in [ad725e2](https://github.com/Kitware/pan3d/pull/33/commits/ad725e2a63525dc5eb16b2779d48f51ec72d9cae)

pan3d/dataset_builder.py
annehaley: The line marking the ui props as dirty was actually leftover from before I had been able to fix the tracking of `expanded_coordinates`. I thought the client was not receiving the updated state, but the state had not been changed at all. Removed this line in [ad725e2](https://github.com/Kitware/pan3d/pull/33/commits/ad725e2a63525dc5eb16b2779d48f51ec72d9cae)

pan3d/dataset_builder.py
annehaley: Removed in [ad725e2](https://github.com/Kitware/pan3d/pull/33/commits/ad725e2a63525dc5eb16b2779d48f51ec72d9cae)

pan3d/dataset_builder.py
annehaley: Changed in [6f497a9](https://github.com/Kitware/pan3d/pull/33/commits/6f497a9be4f3e7e817c6b87aa61a51dcd2ea71bf)

pan3d/dataset_builder.py
annehaley: Changed in [6f497a9](https://github.com/Kitware/pan3d/pull/33/commits/6f497a9be4f3e7e817c6b87aa61a51dcd2ea71bf)

pan3d/dataset_builder.py
```diff
@@ -403,11 +403,14 @@ def mesh_updated(exception=None):
...
-            with open(config_file) as f:
-                config = json.load(f)
-        else:
-            config = json.loads(config_file)
+        if isinstance(config_file, dict):
+            config = config_file
+        elif isinstance(config_file, str):
+            if Path(config_file).exists():
+                with open(config_file) as f:
+                    config = json.load(f)
```
jourdain, line 411: https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text

pan3d/dataset_builder.py
```diff
@@ -485,7 +483,6 @@ def export_config(self, config_file=None):
             config["ui"][state_var] = self.state[state_var]
 
         if config_file:
-            with open(config_file, "w") as f:
-                json.dump(config, f)
+            Path(config_file).open("w").write(json.dumps(config))
```
jourdain, line 481: https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_bytes

pan3d/dataset_builder.py
jourdain: still not what I was thinking for 
https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_text

pan3d/utils.py
jourdain: love it, much easier to grasp what those vars are for.

examples/example_config.json
```diff
@@ -1,12 +1,27 @@
...
-    "state": {
-        "active_array": "analysed_sst",
-        "x_array": "lat",
-        "x_scale": 2,
-        "y_array": "lon",
-        "t_array": "time",
-        "t_index": 5,
-        "main_drawer": false
+    "data_origin": "https://ncsa.osn.xsede.org/Pangeo/pangeo-forge/noaa-coastwatch-geopolar-sst-feedstock/noaa-coastwatch-geopolar-sst.zarr",
+    "data_array": {
```
jourdain, line 3: Thinking more about it. I don't like the name `data_array`. The reason is that it is not a data array but more of a volume or a mesh. 

I want the consistency between `dataset_builder.data_{xxxx}` and that entry `data_{xxxx}` in the config. But I find the current name misleading. 

Should we use `data_mesh` instead? @johnkit

pan3d/dataset_builder.py
```diff
@@ -261,77 +281,74 @@ def on_set_array_active(self, array_active, **kwargs):
...
+        self.plotter.clear()
+        self.plotter.view_isometric()
 
-    @change("x_array")
-    def on_set_x_array(self, x_array, **kwargs):
-        self.algorithm.x = x_array
+    @change("da_x")
+    def on_set_da_x(self, da_x, **kwargs):
+        self.algorithm.x = da_x
         self.mesh_changed()
```
jourdain, line 293: I would group those into a single method.
Something like

```python
@change("da_x", "da_y", "da_z")
def on_coordinate_change(self, da_x, da_y, da_z, **kwargs):
      self.algorithm.x = da_x
      self.algorithm.y = da_y
      self.algorithm.z = da_z
      self.mesh_changed()
```

pan3d/dataset_builder.py
```diff
@@ -261,77 +281,74 @@ def on_set_array_active(self, array_active, **kwargs):
...
-        self.plotter.update()
-
-    @change("x_scale", "y_scale", "z_scale")
-    def on_set_scale(self, x_scale=None, y_scale=None, z_scale=None, **kwargs):
-        x_scale = x_scale or self.state.x_scale
-        y_scale = y_scale or self.state.y_scale
-        z_scale = z_scale or self.state.z_scale
-        self.plotter.set_scale(xscale=x_scale, yscale=y_scale, zscale=z_scale)
+    @change("ui_dialog_shown")
+    def on_set_ui_dialog_shown(self, ui_dialog_shown, **kwargs):
```
jourdain, line 320: `on_dialog_visibility_change(...`

pan3d/ui/coordinate_configure.py
```diff
@@ -39,10 +44,10 @@ def __init__(
                                 v_model=axis_info["index_var"],
                                 min=0,
                                 max=(
-                                    "%s?.size - 1" % coordinate_info,
+                                    f"{coordinate_info}?.size - 1",
                                     0,
                                 ),
-                                step=1,
+                                step=(f"{coordinate_info}?.step", 0),
```
jourdain, line 50: Is the default step supposed to be 0 rather than 1?

pan3d/ui/coordinate_configure.py
```diff
@@ -52,65 +57,83 @@ def __init__(
                                         v_model=axis_info["index_var"],
                                         min=0,
                                         max=(
-                                            "%s?.size - 1" % coordinate_info,
+                                            f"{coordinate_info}?.size - 1",
                                             0,
                                         ),
+                                        step=(f"{coordinate_info}?.step", 0),
```
jourdain, line 63: step of 0 by default?

(might be ok, I just want to double check as from the outside it is strange)

jourdain (changes requested):
getting much better

examples/example_config.json
jourdain: That does not need to be fixed for that PR. It is just a thought that should be addressed before we release.

pan3d/ui/coordinate_configure.py
annehaley: Thanks, changed in [817bc5e](https://github.com/Kitware/pan3d/pull/33/commits/817bc5edce5cb29684f1bf6047e58cce5cf16111)

pan3d/ui/coordinate_configure.py
annehaley: Thanks, changed in [817bc5e](https://github.com/Kitware/pan3d/pull/33/commits/817bc5edce5cb29684f1bf6047e58cce5cf16111)

pan3d/dataset_builder.py
annehaley: I'm hoping to have a consistent naming convention for these state change functions. Right now it's `on_set_[name]`, but if you think it should be `on_[name]_change` I can switch it for all of them, but I do think the function name should include the name of the state variable somewhere.

pan3d/dataset_builder.py
annehaley: That is much more succinct, thank you. Changed in [41e3c22](https://github.com/Kitware/pan3d/pull/33/commits/41e3c224749c308565d13cd4bc51f0e5e62d7423)

pan3d/dataset_builder.py
annehaley: Changed in [b816852](https://github.com/Kitware/pan3d/pull/33/commits/b81685249babc1f58e001ca34c9f7d61bc8db8ea)

pan3d/dataset_builder.py
annehaley: Changed in [b816852](https://github.com/Kitware/pan3d/pull/33/commits/b81685249babc1f58e001ca34c9f7d61bc8db8ea)

pan3d/dataset_builder.py
annehaley: Changed in [b816852](https://github.com/Kitware/pan3d/pull/33/commits/b81685249babc1f58e001ca34c9f7d61bc8db8ea)

pan3d/dataset_builder.py
jourdain: Agree it should be consistent and be like `on_[xxx]_change`, but having the variable name in both the annotation and method name does not seem that useful. Having the intent conveyed by the method name is better. See the grouping comment `on_coordinates_change`

annehaley: I am hesitant to lose the congruence between the annotation and method name. To me it feels messy to refer to the same property by different names (`ui_dialog_shown` and `dialog_visibility`). If the nomenclature of "visibility" is preferred over "shown", I would change it everywhere, not just in the function name.

jourdain: I technically agree, especially if it is a 1-to-1 mapping. But it is still not a good name yet.

Looking at the code, I just realized it is not a boolean but more the name of the dialog you want to show.

What about ui_mode/action? Then a method like `on_mode_change` should be appropriate. No need to repeat the ui_* prefix in the method name. 

Your naming should convey the intent not how you do it (i.e. with a dialog).

pan3d/dataset_builder.py
```diff
@@ -199,43 +210,52 @@ def set_dataset_path(self, dataset_path, **kwargs):
...
+    @change("da_active")
+    def set_da_active(self, da_active, **kwargs):
+        if (
+            da_active is None
+            or not self.state.dataset_ready
+            or da_active == self.da_active
+        ):
             return
+        self.da_active = da_active
+        self.state.da_active = da_active
```
jourdain, line 254: I don't get that line. This is equivalent to `self.state.da_active = self.state.da_active`

Maybe that is why you called those methods set? 

What is the intent of that method? update_active_array?

pan3d/dataset_builder.py
```diff
@@ -261,77 +281,66 @@ def on_set_array_active(self, array_active, **kwargs):
...
-        self.mesh_changed()
+        self.state.dirty("da_coordinates", "ui_expanded_coordinates")
+        self.plotter.clear()
+        self.plotter.view_isometric()
 
-    @change("z_array")
-    def on_set_z_array(self, z_array, **kwargs):
-        self.algorithm.z = z_array
+    @change("da_x", "da_y", "da_z")
+    def on_set_da_axis(self, da_x, da_y, da_z, **kwargs):
```
jourdain, line 291: update_mesh_coordinates?

pan3d/dataset_builder.py
```diff
@@ -261,77 +281,66 @@ def on_set_array_active(self, array_active, **kwargs):
...
+        self.algorithm.z = da_z
         self.mesh_changed()
 
-    @change("t_array")
-    def on_set_t_array(self, t_array, **kwargs):
-        self.algorithm.time = t_array
-        if self.dataset and self.state.array_active and t_array:
-            time_steps = self.dataset[self.state.array_active][t_array]
+    @change("da_t")
+    def on_set_da_t(self, da_t, **kwargs):
```
jourdain, line 298: update_mesh_time?

jourdain: what is da_t vs da_t_index? I can't figure out what each method is doing?

pan3d/dataset_builder.py
jourdain: Side note here. Are those methods aiming to be used by the user or internal?

jourdain: If internal then prefix them with an _. 
If you want the user to use them as well, then make sure they can be understood. Like `set_active_array` or `set_axis_array_names`...

examples/example_config.json
annehaley: It may be worth considering that the dtype of the object returned for `dataset_builder.data_array` is an `xarray.DataArray`. While we typically use it as the source for a volume or mesh, a user could do something else with it if they desire. And the entry in the config includes all the information required to reconstruct that `xarray.DataArray` object.

pan3d/dataset_builder.py
annehaley: The intention of `da_t` is similar to `da_x`, etc. These variables store strings to refer to the name of the array. `da_t` stores the name of the time variable, which would typically be "time", but it could be something like "second" or "day". Whereas `da_t_index` stores an integer which is the current slice along that named axis.

pan3d/dataset_builder.py
annehaley: We could do something similar to what we did for `set_dataset_path`, where the incoming parameter may either be already stored in the state or may be a new change. Like your suggestion for `set_dataset_path`, where we store `self.dataset_path` separately from `self.state.dataset_path` so that we can make the comparison and do a no-op in the case of a match.

pan3d/dataset_builder.py
annehaley: This was my attempt to mimic the suggestion you had made for `set_dataset_path`, where the method can be either called as a state change effect or called directly with a new value. When called in the latter case, this line is necessary. But a redundant call is avoided with the check just above it. `self.da_active` is stored separately from `self.state.da_active` just like you had suggested for `self.dataset_path` being separate from `self.state.dataset_path`. It is mostly redundant storage, but it allows us to check if the side effect has already been called for that value.

pan3d/dataset_builder.py
jourdain: Thanks for the clarification

pan3d/dataset_builder.py
jourdain: So should the method name be `set_dataset_axis_names`?

pan3d/dataset_builder.py
jourdain: That makes sense, especially if that method is not called by the @change. 
Another option (if that makes better sense) would be to do something like

```python
def set_active_array_name(array_name):
    self.state.da_active = array_name

@change("da_active")
def _on_active_array_change(da_active, **kwargs):
    ...
```

pan3d/dataset_builder.py
annehaley: Sure, I like that naming. I can apply that pattern to all the state var side-effect functions so that any of them can be either called by the user directly or as a response to a state change.

pan3d/dataset_builder.py
annehaley: I do like this separation of a user-accessible functions from the state change side effects. It makes it no longer necessary to have redundant storage like `self.dataset_path` versus `self.state.dataset_path`. I'll apply this pattern to all state changes that should be user-accessible.

annehaley: I am trying this out now, and I'm making an example notebook which uses these user-accessible functions back-to-back, like so:
```
builder = DatasetBuilder()

builder.set_dataset_path('../example_dataset.nc')
builder.set_data_array_active_name('density')
builder.set_data_array_axis_names(
    x='length',
    y='width',
    z='height',
    t='seconds',
)
builder.set_data_array_time_index(2)
```
But this breaks because the state change callbacks are not called in-between. The callback for when `dataset_path` is set needs to occur before the callback after `da_active` is set. Since we cannot enforce callback order on the final flush, Is there a way that we can cause a flush to occur in these user-accessible functions?

jourdain: Probably yes, but I would not advise it. 

Just keep that specific one that need explicit execution order the way you have it.

annehaley: Maybe I can try a combination of the techniques. The user-accessible function could call the private function, and the private function can still use a check against `self.state.dataset_path` equaling `self.dataset_path`? This will avoid performing the side effects twice.

jourdain (approved):

PR #34, opened 2023-10-05
Loading optimization

annehaley:
When using remote datasets (like the examples from Pangeo Forge), the load time has become too long ever since we switched to more specific slicing along our axes.

Currently, we fetch the whole data array (for the given time slice) and retroactively apply our start-stop-step slicing to x, y, and z. This is because the Pyvista-Xarray algorithm class had no support for this type of slicing. 

This PR, in tandem with https://github.com/pyvista/pyvista-xarray/pull/58, eliminates most of that load time. See the below diagram for load time results. There are three groups of five load times in this diagram:

1. No fix applied (code from `main` with some extra print statements to measure times)
2. Partial fix applied (First commit from https://github.com/pyvista/pyvista-xarray/pull/58 only)
3. Full fix applied, with data persisting and `isel` indexing

image

(This PR also has some initial commits which fix small things I noticed during load time testing, as well as a new example file.)


examples/example_config_cmip.json
```diff
@@ -0,0 +1,42 @@
...
+    "data_origin": "https://ncsa.osn.xsede.org/Pangeo/pangeo-forge/cmip6-feedstock/test_full_depth.zarr",
+    "data_array": {
+        "active": "so",
+        "x": "i",
+        "y": "j",
+        "z": "lev",
+        "t": "time"
+    },
+    "data_slices": {
+        "time": [
```
alesgenova, line 11: Ok, I can see why you used a plain `list` instead of `slice` in the other PR

pan3d/dataset_builder.py
```diff
@@ -364,24 +348,26 @@ def auto_select_coordinates(self):
...
-
-            self.state.ui_unapplied_changes = True
+        self.da = self.algorithm.sliced_data_array
+        if self.state.da_active and self.da is not None:
+            total_bytes = self.da.size * self.da.dtype.itemsize
+        else:
+            total_bytes = 0
+        exponents_map = {0: "bytes", 1: "KB", 2: "MB", 3: "GB"}
+        for exponent in sorted(exponents_map.keys(), reverse=True):
+            divisor = 1024**exponent
```
alesgenova, line 358: Most likely unnecessary optimization:
- bytes: `1 << 0`
- KB: `1 << 10`
- MB: `1 << 20`
- GB: `1 << 30`

alesgenova (approved):
Looks good!

alesgenova (approved):
Looks good, the code just need to be re-linted to make the CI pass

alesgenova (approved):

PR #35, opened 2023-10-11
Update PyPI publishing

annehaley:
- Python semantic release has changed their API as of version 8 to no longer offer PyPI publishing. This PR follows their migration guide. https://python-semantic-release.readthedocs.io/en/latest/migrating_from_v7.html
- PyPI publishing is now done as a separate step using `pypa/gh-action-pypi-publish`, as recommended by the PyPI docs: https://docs.pypi.org/trusted-publishers/using-a-publisher/
- I have created a pending Trusted Publisher on my PyPI account, which is waiting for the first push from this repository to create the PyPI project.

### DIFF
--- a/.github/workflows/test_and_release.yml
+++ b/.github/workflows/test_and_release.yml
@@ -69,18 +69,22 @@ jobs:
       pytest,
     ]
     runs-on: ubuntu-latest
-    if: github.event_name == 'push'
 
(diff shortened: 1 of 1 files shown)


jourdain (approved):
I guess we'll know if it works when we merge ;-)

PR #36, opened 2023-11-06
Add UI panel to change render options

annehaley:
New render options include:

- colormap
- transparency
- transparency function
- warp by scalars
- x scale
- y scale
- z scale

image

(Additionally, the colormap scale key is interactive; it can be resized and moved. This is functionality built into Pyvista.)

### DIFF
--- a/pan3d/dataset_builder.py
+++ b/pan3d/dataset_builder.py
@@ -13,7 +13,7 @@ from trame.widgets import html, client
 from trame.widgets import vuetify3 as vuetify
 
 from pan3d.pangeo_forge import get_catalog
-from pan3d.ui import AxisDrawer, MainDrawer, Toolbar
+from pan3d.ui import AxisDrawer, MainDrawer, Toolbar, RenderOptions
 from pan3d.utils import initial_state, run_singleton_task, coordinate_auto_selection
 
 BASE_DIR = Path(__file__).parent
@@ -94,6 +94,7 @@ class DatasetBuilder:
                         v_show=("ui_error_message",),
                     )
                     with html.Div(style="height: 100%; position: relative"):
+                        RenderOptions()
                         with plotter_ui(
                             self.ctrl.get_plotter(),
                             interactive_ratio=1,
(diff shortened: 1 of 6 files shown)


pan3d/dataset_builder.py
```diff
@@ -367,11 +430,23 @@ def mesh_changed(self):
...
+
+        mesh = self.mesh
+        if self.state.render_scalar_warp:
+            mesh = mesh.warp_by_scalar()
+        self.actor = self.plotter.add_mesh(
+            mesh,
+            **args,
         )
         self.plotter.view_isometric()
+        self.plotter.reset_camera()
```
jourdain, line 449: Did that part fix the local rendering issue?

jourdain (dismissed):

pan3d/dataset_builder.py
annehaley: Sort of; changing the colormap or transparency is a smooth transition now, but changing "warp by scalars" or the x-y-z scales still results in the mesh appearing to move away from the center (or disappear entirely if the change was drastic).

jourdain: So it does not work. Since only the last 2 are what require the camera to be updated.
I don't think it is an API on the plotter that you need to call. This is something on the viewer. And it could be like a push_camera or something. You can use the controller to get your hand on that function.

annehaley: Hm, ok. I have now tried using `self.ctrl.reset_camera()` and `self.ctrl.view_update()`, with different combinations, and I'm able to get the "warp by scalars" change to transistion smoothly with [d2f8baf](https://github.com/Kitware/pan3d/pull/36/commits/d2f8baf773c9a5e68836f96cc45348f1ce20ed0a), but the x-y-z scale changes are behaving the same.

jourdain: You may have to do 


```
self.ctrl.reset_camera() # this line should be the one not needed
self.ctrl.view_update()
self.ctrl.reset_camera()
```

jourdain: Not sure but I think the `view_update` can take a `reset_camera=True` arg

annehaley: It does take that flag, and I'm trying different combinations of the two lines (also switching the order as you suggested), but I am getting the same behavior on these permutations.

annehaley:
@jourdain Sorry, can you leave your review again? The previous one was dismissed after I merged #34

jourdain (approved):

PR #37, opened 2023-11-08
Pangeo datasets compatibility

annehaley:
Going through the list of datasets available via Pangeo Forge, some datasets were incompatible due to their time indexing. This PR makes Pan3D compatible with those datasets by changing how we handle time indexing. 

This PR also makes some UI changes that became necessary when viewing these Pangeo Forge datasets:

- [x] Show the current time stamp as a formatted string above the time slider
- [x] Move the data attributes table to a dialog. The Pangeo Forge datasets have a lot of this metadata, and viewing it in the narrow side panel was frustrating.
- [x] Allow the user to view attributes tables for each available array (same dialog format as Dataset attributes). The Pangeo Forge datasets have a lot of metadata stored at this level, too.


pan3d/dataset_builder.py
```diff
@@ -70,6 +71,12 @@ def ctrl(self):
     def data_array(self):
         return self.algorithm.sliced_data_array
 
+    @property
+    def pv_mesh(self):
```
jourdain, line 75: is pv_mesh a ParaView proxy, or pyvista mesh or a vtk dataset?

jourdain (approved):
Looks reasonable to me

pan3d/dataset_builder.py
jourdain: I might prefer a `vtk_mesh` name

annehaley:
Thanks for the review @jourdain! It looks like the protections on this repo only allow us to use one active approval at a time. Could you leave your review again? (And do you think I should turn off this "dismiss stale reviews" feature for main?)

jourdain (dismissed):

pan3d/dataset_builder.py
annehaley: The object returned from this is a `pyvista.core.grid.RectilinearGrid`

jourdain: then name it `rectilinear_grid`

annehaley: Apologies, it could also be a `pyvista.core.grid.StructuredGrid`, according to this logic in `pyvista-xarray`: https://github.com/pyvista/pyvista-xarray/blob/ff112c0979147a085cffe2e82790c039f829e402/pvxarray/accessor.py#L74. Maybe the name could be `pv_grid` or `pyvista_grid`?

jourdain: I would go with either `grid`, `mesh`, `vtk`.

annehaley: Ok, I like the thought of still calling it a mesh, since that's what we call it internally. I changed the name of that property to just "mesh" and added an underscore to all previous instances of "self.mesh": [75bd691](https://github.com/Kitware/pan3d/pull/37/commits/75bd691633ff50bc43447f25c5c53d2475e6cdcd)

jourdain (approved):

PR #38, opened 2023-11-16
Fix PyPI publishing step in CI

annehaley:
The CI config in #35 failed with the following:
```
The workflow is not valid. .github/workflows/test_and_release.yml (Line: 88, Col: 9): Unexpected value 'environment' .github/workflows/test_and_release.yml (Line: 89, Col: 9): Unexpected value 'permissions'
```

This PR changes the yml file by separating the PyPI publish step as its own job after the semantic release job. This time the job definition was pasted directly from the PyPI documentation.

### DIFF
--- a/.github/workflows/test_and_release.yml
+++ b/.github/workflows/test_and_release.yml
@@ -81,10 +81,20 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


jourdain (approved):

jourdain:
Now that you have CI with semantic release, you will need to start using proper format for your commit messages.

PR #39, opened 2023-11-16
Fix release job

annehaley:
With #38, the semantic release step completed, but the following problems were identified:

- Version was set to 0.0.0, which means the version specification in `setup.cfg` was ignored. The cause was identified in the semantic-versioning v8 breaking changes. Configuration in `setup.cfg` is no longer supported (https://python-semantic-release.readthedocs.io/en/latest/migrating_from_v7.html#setup-cfg-is-no-longer-supported). This PR adds a pyproject.toml file, which contains the same information as `setup.cfg`. Should we remove `setup.cfg`?
- The `CHANGELOG.md` file, which was pushed with the 0.0.0 release, contains all commit messages back to the first. One of those had a typo which caused a codespell error. This PR fixes the typo.

The PyPI release job that followed failed with the following:
```
Warning:  It looks like there are no Python distribution packages to publish in the directory 'dist/'. Please verify that they are in place should you face this problem.
ERROR    InvalidDistribution: Cannot find file (or expand pattern): 'dist/*'  
```

This PR adds a job to the release workflow to build the dist folder prior to the pypi publish job.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,51 @@
+[project]
+name = "pan3d"
+version = "1.0.0"
+description = "Utility package for processing and visualizing 3D datasets"
+authors = [
+    {name = "Kitware Inc."},
+]
+dependencies = [
+    "dask",
+    "fsspec",
+    "matplotlib",
+    "netCDF4",
+    "pyvista",
+    "pyvista-xarray",
+    "trame",
+    "trame-vtk",
+    "trame-vuetify",
+    "vtk",
+    "xarray",
+    "zarr",
+]
+requires-python = ">=3"
+readme = "README.rst"
+license = {text = "Apache Software License"}
+keywords = ["Python", "Interactive", "Web", "Application", "Framework"]
+classifiers = [
+    "Development Status :: 4 - Beta",
+    "Environment :: Web Environment",
+    "License :: OSI Approved :: Apache Software License",
+    "Natural Language :: English",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3 :: Only",
+    "Topic :: Software Development :: Libraries :: Application Frameworks",
+    "Topic :: Software Development :: Libraries :: Python Modules",
+]
+
+[project.scripts]
+pan3d-viewer = "pan3d.__main__:main"
+
+[project.entry-points.jupyter_serverproxy_servers]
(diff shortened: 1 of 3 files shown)


jourdain:
Do you know that a `fix` trigger a release while most of the work you do is `ci:` related and should not trigger a new release.

pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+    "pyvista",
+    "pyvista-xarray",
+    "trame",
+    "trame-vtk",
+    "trame-vuetify",
+    "vtk",
+    "xarray",
+    "zarr",
+]
+requires-python = ">=3"
```
jourdain, line 22: 3.7?

jourdain (approved):
Since you are testing the ci, I'm ok with your "fix" commit message but be aware that the purpose of semantic release is not to trigger a new release when no code change happened.

annehaley:
Ok, thanks. I hadn't known about that side effect. I will keep these because we do want to run the release job in this instance, but I will keep that in mind for future PRs. Does `feat` have the same side effect?

pyproject.toml
annehaley: Changed in [bffff73](https://github.com/Kitware/pan3d/pull/39/commits/bffff73681a051707ed5a100e8e86bebec4a030b)

jourdain:
What you call "side effect" is the actual intent of semantic-release. 

```
fix: +0.0.1
feat: +0.1
BREAKING CHANGE: +1 (Last line of commit message not in subject)
```

See https://github.com/angular/angular/blob/main/CONTRIBUTING.md#-commit-message-format

jourdain:
setup.py
```python
from setuptools import setup

setup()
```

annehaley:
My mistake, I had typed `setup.py` where I meant `setup.cfg`. The `cfg` file is what I used to write the `pyproject.toml`, and it is now redundant. Unless there are some cases where that file is needed?

jourdain:
Yes the `pyproject.toml` is the modern version of the `setup.cfg`. You should remove `setup.cfg` now that you have a `pyproject.toml`.

annehaley:
Thanks, I removed `setup.cfg` in [a59ef67](https://github.com/Kitware/pan3d/pull/39/commits/a59ef6764bde3426dae90feee312d18fc22c6927)

jourdain (approved):

PR #40, opened 2023-11-20
Packaging

annehaley:


jourdain (approved):

README.rst
```diff
@@ -22,7 +22,7 @@ Run the application
 
 .. code-block:: console
 
-    pan3d-viewer --data /path/to/directory.zarr
+    pan3d-viewer --dataset_path /path/to/directory.zarr
```
jourdain, line 25: weird arg name. Should we use `--dataset` instead?

annehaley: Ok sure, should we change the name throughout the DatasetBuilder class and in the import/export format? Or just for the entrypoint argument?

jourdain: not sure then. It is just weird to have `_` in a var of the cli

annehaley: Ok, we'll just change the CLI arg and continue to distinguish between `dataset_path` and `dataset` internally to the `DatasetBuilder`. I made the change in [35fb509](https://github.com/Kitware/pan3d/pull/40/commits/35fb509cd7b63a1bec54927c85597eca20727410).

PR #42, opened 2023-12-04
Add `vtk-osmesa` to requirements for binder

annehaley:
Currently, running the example notebooks on Binder gives the following error:
image

This may be due to the fact that we should install `vtk-osmesa` in the binder kernel, so this PR makes that addition to `examples/jupyter/requirements.txt`.

### DIFF
--- a/examples/jupyter/Dockerfile
+++ b/examples/jupyter/Dockerfile
@@ -1,0 +1,7 @@
+FROM ghcr.io/pyvista/pyvista:latest
+
+COPY . ${HOME}
+WORKDIR ${HOME}
+RUN pip install -r requirements.txt
+RUN pip uninstall vtk -y
+RUN pip install --extra-index-url https://wheels.vtk.org vtk-osmesa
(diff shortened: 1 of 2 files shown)


jourdain:
Yes osmesa is required in that context. But does it work now? Just worried that we have 2 vtk rather than just 1.

annehaley:
It works in my local context, but I do see that Bane warns about a namespace resolution error in this announcement: https://discourse.vtk.org/t/status-update-vtk-python-wheels/11212. 
Perhaps we can do what he does in the pyvista-binder cookiecutter project? This example uses a Dockerfile which inherits from a pyvista base image: https://github.com/pyvista/cookiecutter-pyvista-binder/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/Dockerfile
I know that Binder warns against using Dockerfile configuration unless it's absolutely necessary, but maybe this is one of those cases.

jourdain:
if it works we can call it the day. It is just technically we should only have 1 vtk. 

Maybe we do `pip install pan3d[osmesa]` to make sure VTK is not a default dependency.

jourdain (approved):

jourdain:
Where is the `requirements.txt` file?

jourdain:
Ideally it should contain `trame-jupyter-extension`

PR #43, opened 2023-12-05
Fix build versioning

annehaley:
Currently, the build action always generates a tar and wheel for version 0.0.0. This results in a failure of the PyPI publish step because v0.0.0 already exists. This PR fixes the build step by adding `setuptools_scm` to `pyproject.toml`. The result should be that PyPI is updated with a version >0.0.0.

### DIFF
--- a/.github/workflows/test_and_release.yml
+++ b/.github/workflows/test_and_release.yml
@@ -63,32 +63,21 @@ jobs:
         pytest -s ./tests
       working-directory: .
 
-  build:
-    name: Build distribution
-    runs-on: ubuntu-latest
-
-    steps:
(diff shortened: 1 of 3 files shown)


jourdain (approved):
I don't know the latest process with semantic release so I trust you.

But the way it was working before was that semantic-release will update the version in the code before the publish along with the changelog. So you will always get new commit on gh after a release.

annehaley:
Ok, I combined the jobs into one in [43bfff8](https://github.com/Kitware/pan3d/pull/43/commits/43bfff81ca595b69aaadc2e287ba2bdda9041cb6). Is this what you had in mind?

jourdain (approved):

PR #45, opened 2023-12-06
Binder configuration

annehaley:
This PR adds a `/binder` folder for the binder configuration files: `requirements.txt` for python requirements, `apt.txt` for system requirements, and `start`, an entrypoint script pasted from the pyvista docs: https://docs.pyvista.org/version/stable/getting-started/installation.html#running-on-remote-servers


binder/start
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -x
+export DISPLAY=:99.0
+export PYVISTA_OFF_SCREEN=true
+which Xvfb
```
zachmullen, line 5:
```suggestion
```

zachmullen (changes requested):

binder/apt.txt
```diff
@@ -0,0 +1,6 @@
+libgl1
+libgl1-mesa-glx
+libosmesa6-dev
+gcc
+xvfb
+python3-opencv
```
jourdain, line 6: what for?

binder/start
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -x
+export DISPLAY=:99.0
+export PYVISTA_OFF_SCREEN=true
+Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
+sleep 3
+set +x
+exec "$@"
```
jourdain, line 8: none of the above is needed

binder/apt.txt
```diff
@@ -0,0 +1,6 @@
+libgl1
+libgl1-mesa-glx
+libosmesa6-dev
+gcc
+xvfb
```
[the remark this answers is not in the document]
annehaley: Without xvfb, I got the following error when running a notebook:
```
This system does not appear to be running an xserver.
PyVista will likely segfault when rendering.

Try starting a virtual frame buffer with xvfb, or using
  `pyvista.start_xvfb()`
```

The Pyvista docs linked above include xvfb in their suggested configuration.

binder/start
annehaley: This script came from the pyvista docs, and I was getting the error mentioned above without this script.

binder/apt.txt
annehaley: Without opencv, I was gettting the following error:  `ImportError: libGL.so.1: cannot open shared object file: No such file or directory`

jourdain: hum, ok. GL is definitely a subset of opencv.

binder/start
jourdain: Yes that is when you use regular vtk not the osmesa one.

annehaley: I'm including `vtk-osmesa` in `requirements.txt`, but I'm not able to get it working without xvfb and the start script. Could you show me what you mean?

jourdain: if xvfb is needed, that is because vtk-osmesa is not the one that is running. It is the default vtk that need X.
Also vtk and vtk-osmesa technically are the same... So if you don't remove vtk before installing vtk-osmesa, who knows what you are getting since they write the same files/directories...

zachmullen: If this is the case, then it should be the case that we can just `pip install vtk` rather than `vtk-osmesa`, and it should still work, right?

jourdain: well if it working now then you are all-set since osmesa is not used.
But the point of osmesa was to skip xvfb and other system stuff

jourdain: BTW, `vtk` is a dependency of `pyvista` so we don't even do `pip install vtk`

zachmullen:
Did we try yet to just not install `vtk-osmesa` and use Xvfb? It seemed like that was working, so probably the easier option.

annehaley:
In both local environments (docker and native) and on binder, any combination that uninstalls the original vtk and then installs vtk-osmesa gives the same error. Perhaps the linkage between pyvista and vtk-osmesa is broken and I should file an issue? 
For now, I have returned to the `xvfb` configuration that pyvista currently recommends.

jourdain:
I think we should report the compatibility issue on PyVista with the vtk-osmesa piece.

jourdain (approved):

PR #46, opened 2023-12-11
More binder configuration improvements

annehaley:
Binder link: https://mybinder.org/v2/gh/Kitware/pan3d/binder-test?labpath=examples%2Fjupyter

**Note:** for the purposes of trying out the binder configuration with current changes to the pan3d module, the binder setup now requires the local version of pan3d with `.` instead of the current version available on PyPI with `pan3d`. We can decide to undo this change before merging.


jourdain (approved):

annehaley:
@jourdain Before merging, do you think I should remove the change mentioned in the description note? For testing, we use a relative path to install pan3d in binder. Going forward, should we keep this change or use the latest available through PyPI?

jourdain:
I was indeed wondering. But I think keeping the `.` is fine. Maybe just add a comment with the actual package name on pypi.

jourdain (approved):

PR #48, opened 2023-12-14
Client rendering fixes

annehaley:
This PR fixes the following errors we had been seeing:

1. `Uncaught TypeError: Cannot mix BigInt and other types, use explicit conversions`

- [x] Solution: require `trame-vtk>=2.6.3`. This version includes a fix by @jourdain to convert dtypes before meshing.

2. Initial camera reset not working in cloud environments where we set default rendering mode to "client"

- [x] Solution: use `plotter.push_camera` instead of `plotter.reset_camera` when cloud mode is enabled

### DIFF
--- a/pan3d/dataset_builder.py
+++ b/pan3d/dataset_builder.py
@@ -505,10 +505,7 @@ class DatasetBuilder:
             **args,
         )
         self.plotter.view_isometric()
-        if self._cloud:
-            self.ctrl.push_camera()
-        else:
-            self.ctrl.reset_camera()
+        self.ctrl.push_camera()
         self.ctrl.view_update()
 
     def reset(self, **kwargs):
--- a/pan3d/dataset_builder.py
+++ b/pan3d/dataset_builder.py
@@ -37,6 +37,7 @@ class DatasetBuilder:
         server.client_type = "vue3"
         self.server = server
         self._layout = None
+        self._cloud = is_cloud_env()
 
         self.state.update(initial_state)
         self.algorithm = PyVistaXarraySource()
@@ -60,7 +61,7 @@ class DatasetBuilder:
         if state:
             self.state.update(state)
 
-        if is_cloud_env():
+        if self._cloud:
             pyvista.global_theme.trame.default_mode = "client"
 
     # -----------------------------------------------------
@@ -117,6 +118,7 @@ class DatasetBuilder:
                         ) as plot_view:
                             self.ctrl.view_update = plot_view.update
                             self.ctrl.reset_camera = plot_view.reset_camera
+                            self.ctrl.push_camera = plot_view.push_camera
         return self._layout
 
     # -----------------------------------------------------
@@ -370,7 +372,7 @@ class DatasetBuilder:
         if self.state.render_scalar_warp != scalar_warp:
             self.state.render_scalar_warp = scalar_warp
 
-        if self._mesh:
+        if self._mesh is not None and self.data_array is not None:
             self.plot_mesh()
 
     # -----------------------------------------------------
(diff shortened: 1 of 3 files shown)


pan3d/dataset_builder.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, server=None, dataset_path=None, state=None, pangeo=False):
         server.client_type = "vue3"
         self.server = server
         self._layout = None
+        self._cloud = is_cloud_env()
```
jourdain, line 40: I'm not keen on that "cloud" name. My issue with that is that it does not reflect your intent. 
You are relying on an assumption that "cloud" means no GPU, and therefore, you should try to prevent server-side rendering. 

Instead, I would pick names like `self._force_local_rendering = not has_rendering_capabilities()` or something else but along those lines that explain your intent.

pan3d/dataset_builder.py
```diff
@@ -503,12 +505,17 @@ def plot_mesh(self):
             **args,
         )
         self.plotter.view_isometric()
-        self.ctrl.reset_camera()
+        if self._cloud:
+            self.ctrl.push_camera()
+        else:
+            self.ctrl.reset_camera()
         self.ctrl.view_update()
```
jourdain, line 512: I don't think you need to do a if. You can make the same call for both. You should validate that assumption, but I would rather have a unified code base than a "complex" if for managing different strategies.

jourdain (changes requested):
Just small details, but now that we are getting close, it is easier to tackle the details. ;-)

pan3d/dataset_builder.py
annehaley: Ok, gotcha. I confirmed that it functions properly with the default rendering mode and made the change in [8f963ff](https://github.com/Kitware/pan3d/pull/48/commits/8f963ff9c11a68d51ffbf87a287ecb88eed754a6).

jourdain (changes requested):
self._cloud = is_cloud_env()

annehaley:
> self._cloud = is_cloud_env()

Should we not store this variable on the builder?

jourdain:
It is fine to store it, but it should reflect your intent.

jourdain:
See my initial comment on that section

annehaley:
Oh thanks. I had to refresh to see that initial comment. I'll make that change.

pan3d/dataset_builder.py
annehaley: Changed in [033505b](https://github.com/Kitware/pan3d/pull/48/commits/033505bd41f1a3067cf3c42cc846c271ea3baaa5)

pan3d/utils.py
```diff
@@ -51,9 +51,10 @@ async def coroutine():
     singleton_task = asyncio.run_coroutine_threadsafe(coroutine(), task_loop)
 
 
-def is_cloud_env():
-    cloud_env_vars = ["BINDER_REQUEST"]
-    return any(os.environ.get(k) for k in cloud_env_vars)
+def force_local_rendering():
```
jourdain, line 54: This method does not `force_local_rendering`. If I'm not mistaken, the intent of that method it to check if you have a GPU which then get used for something else such as "should I force client side rendering".

pan3d/dataset_builder.py
jourdain: Please read my comment fully

pan3d/utils.py
annehaley: I see the distinction you mean. How about `has_gpu_rendering`?

annehaley: Change made in [094c3bc](https://github.com/Kitware/pan3d/pull/48/commits/094c3bc80ec74777fbd3ec35a92a6299f575a126)

jourdain (approved):

PR #70, opened 2024-03-18
Validate inputs to DatasetBuilder setters

### LINKED ISSUE #49: Fortify `DatasetBuilder` methods against unexpected inputs
After writing documentation in #47, users will be more likely to call methods on the `DatasetBuilder` class manually. We will eventually need to bolster these methods with more robust input handling that will raise informative errors when unexpected values are received.

annehaley:
Resolves #49

### DIFF
--- a/pan3d/dataset_builder.py
+++ b/pan3d/dataset_builder.py
@@ -148,8 +148,8 @@ class DatasetBuilder:
                 raise ValueError(
                     "Cannot set data array name without setting dataset info first."
                 )
-            acceptable_values = list(self.dataset.data_vars.keys())
-            if data_array_name not in acceptable_values:
+            if data_array_name not in self.dataset.data_vars:
+                acceptable_values = list(self.dataset.data_vars.keys())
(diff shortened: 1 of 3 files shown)


pan3d/dataset_builder.py
```diff
@@ -124,6 +141,18 @@ def data_array_name(self) -> Optional[str]:
...
     def data_array_name(self, data_array_name: Optional[str]) -> None:
+        if data_array_name is not None:
+            if not isinstance(data_array_name, str):
+                raise TypeError("Type of data_array_name must be str or None.")
+            if self.dataset is None:
+                raise ValueError(
+                    "Cannot set data array name without setting dataset info first."
+                )
+            acceptable_values = list(self.dataset.data_vars.keys())
+            if data_array_name not in acceptable_values:
```
jourdain, line 152: Why not just `if data_array_name not in self.dataset.data_vars:`? And then generate the keys in case of error...

jourdain (approved):

PR #77, opened 2024-04-19
Create pan3d-cloud docker image

annehaley:
Images `kitware/trame:pan3d-cloud` and `kitware/trame:pan3d-geotrame` are now available at https://hub.docker.com/r/kitware/trame/tags


.github/workflows/docker_image.yml
```diff
@@ -1,38 +1,47 @@
...
+          password: ${{ secrets.DOCKER_TOKEN }}
       - name: Build and push the Docker image
         uses: docker/build-push-action@v3
         with:
           context: .
           file: docker/Dockerfile
           push: ${{ github.actor != 'dependabot[bot]' }}
-          tags: ${{ steps.meta.outputs.tags }}
-          labels: ${{ steps.meta.outputs.labels }}
+          tags: kitware/geotrame-viewer
```
jourdain, line 28: We don't have a dockerhub repository here... It needs to be `kitware/trame:geotrame-viewer`

.github/workflows/docker_image.yml
```diff
@@ -1,38 +1,47 @@
...
+        with:
+            username: ${{ secrets.DOCKER_USERNAME }}
+            password: ${{ secrets.DOCKER_TOKEN }}
+      - name: Build and push the Docker image
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          file: docker/cloud.Dockerfile
+          push: ${{ github.actor != 'dependabot[bot]' }}
+          tags: kitware/geotrame-cloud
```
jourdain, line 47: wrong tag name

annehaley: Changed in [3e3251d](https://github.com/Kitware/geo-trame/pull/77/commits/3e3251d2c0b1196a2cfbd2a34d9871051439bbea)

.github/workflows/docker_image.yml
annehaley: Changed in [3e3251d](https://github.com/Kitware/geo-trame/pull/77/commits/3e3251d2c0b1196a2cfbd2a34d9871051439bbea)

docker/cloud.Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM pangeo/pangeo-notebook:latest
+
+RUN pip install pan3d[all] imageio trame-jupyter-extension
+RUN pip uninstall -y vtk
+RUN pip install --extra-index-url https://wheels.vtk.org vtk-osmesa
+
+COPY ./examples /shared/pan3d_examples
```
johnkit, line 7: Just a sanity check: Is `/shared` the location that Jim Colliander told us to put our shared files? Seems right, but I don't have anything in my notes from our last meeting with him.

annehaley: I don't remember Jim specifying anything about the file locations, but Yuvi had stated in his last email that "if you have something in the docker *image*, my suggestion is to put that somewhere other than `$HOME` (like `/srv/shared`), and that would automatically be accessible to all users. This is because just $HOME is overriden by persistent storage, but everything else works as normal." 

Maybe we could try `/srv/shared` as he suggests, but by this description `/shared` should work. However, the behavior we observed is that this location is overwritten in the way he says the `$HOME` location is. Not sure what location is best to make the examples available to all users by default (and showing up in their file tree). In my recollection, we haven't gotten a straight answer about that.

annehaley:
The latest repsonse we received from Yuvi at 2i2c included an explanation that the extra "pan3d-viewer" tile on the Jupyter launch page was due to the configuration in `pan3d/jupyter.py`. After conferring with @jourdain, I removed this file since that configuration is no longer applicable (a remnant from an earlier version of the project)

johnkit (approved):

jourdain (approved):

PR #78, opened 2024-05-02
Improve Coordinate Indexing & Slicing

annehaley:
This PR includes the following changes:

- [x] Internal structure for coordinate slicing now uses indices instead of true values for start and stop. This permits the user to use non-numeric coordinates (such as time) as a spatial coordinate.
- [x] Add a `resolution` argument to the `DatasetBuilder` (also available as a CLI arg). If not specified, `resolution` defaults to 128. This value is used for auto-slicing, where the step value is inferred from the start and stop values and the resolution. If the user specifies a value <=1, auto-slicing will be disabled. If auto-slicing is disabled, the user may edit start, stop, and step in the Axis Drawer (previous implementation).
- [x] If auto-slicing is enabled, the inputs for start, stop, step in the Axis Drawer are no longer available to the user, but a new UI component is available in the rendering area. The "Configure Bounds" menu, when open, displays a set of sliders to change the bounds (start and stop) of each coordinate axis (figure 1 below). There is also an option to enable an interactive preview mode (figure 2 below).
- [x] When enabled, the Interactive Preview mode displays a custom widget implemented directly with Vue (rather than using existing Trame widgets). This introduces some new code structures in the `pan3d/ui/vue` folder. This widget displays a preview image of one face of the rendered 3d structure and uses SVG elements to allow the user to resize a selection box which changes the bounds of the two coordinates along that face. The user may select which face to preview, and clicking the camera button next to this input will snap the camera to view that face. When this widget is shown, the two sliders which represent coordinates changed by the selection box will disappear. The time slider and the slider for the axis normal to the selected face will still be available. 
- [ ] **TODO**: Documentation needs updating, including tutorial screenshots (this can be done in a future PR when we update all docs with the new name)

image
image


johnkit:
I went through alot of the visualization features and everything works fine for me. There were a couple things that did not work as expected, though they might not be due to this PR (I didn't go back and compare with pan3d:main branch).

* The "Import" feature does not work for me. When I click the "Import" button, the viewer displays the "Import File Select" dialog. I can browse to my config file ok. When I click the dialog's "Import" button, the dialog remains displayed and the view does not update. (No messages on the terminal or js console.) Here's the config file I tested with: 
[pan3d_state.json](https://github.com/Kitware/geo-trame/files/15470116/pan3d_state.json)


* This might be more of a feature request: When the user switches to cartography mode, I think we should change the center of rotation to the earth's center.

pan3d/ui/vue/.eslintrc.cjs
jourdain, line 1: Why do you nest JS source inside the Python package?

annehaley: Just for semantic proximity. I figured it belongs in the UI folder where we define all our components, but it should have its own subfolder `vue`, where we include the JS source and the module definition which refers to that source. The python component class is grouped with the other python component classes, though. Do you have a different organization in mind?

jourdain: you can use `pan3d-js/` as root directory to store the JS project structure. Then when the project build make it write in the `serve` directory of the Python part...

pan3d/ui/pan3d_components/module/__init__.py
```diff
@@ -0,0 +1,13 @@
+from pathlib import Path
+
+# Compute local path to serve
+serve_path = str(Path(__file__).parent.with_name("serve").resolve())
```
jourdain, line 4: serve should be nested into the module. Otherwise it is not a module if it needs files outside of its tree...

pan3d/ui/preview_bounds.py
```diff
@@ -0,0 +1,14 @@
+from trame_client.widgets.core import HtmlElement
```
jourdain, line 1: This file should be in `pan3d/ui/widgets.py`

pan3d/ui/pan3d_components/module/__init__.py
annehaley: I copied this code from https://github.com/Kitware/trame-components/blob/master/trame_components/module/__init__.py, so I'm not clear on what its intent is. Could you clarify what change you mean?

jourdain: Should be with the following path `pan3d/ui/module/`.

pan3d/ui/preview_bounds.py
annehaley: Do you mean just renaming `preview_bounds` -> `widgets`? Or should it be moved as well?

jourdain: just rename it.

pan3d/ui/pan3d_components/module/__init__.py
jourdain: No you did not, you've inserted `.parent`

pan3d/ui/pan3d_components/module/__init__.py
```diff
@@ -1,7 +1,7 @@
 from pathlib import Path
 
 # Compute local path to serve
-serve_path = str(Path(__file__).parent.with_name("serve").resolve())
+serve_path = str(Path('pan3d/ui/pan3d_components/module'))
```
jourdain, line 4: __Old__ `str(Path(__file__).parent.with_name("serve").resolve())`
__New__ `str(Path(__file__).with_name("serve").resolve())`

jourdain: The module directory should contains all it needs.

annehaley: Do you mean that I should move the `serve` directory within the `module` directory? Otherwise this new line results in a "no directory exists" error.

jourdain: yes, you need to fix both for consistency.

annehaley: Ok, changed in [06bc8bd](https://github.com/Kitware/geo-trame/pull/78/commits/06bc8bd0840bba4f662dbd1d16ac0d9180d69190)

.gitignore
```diff
@@ -4,7 +4,7 @@ node_modules
 
 examples/jupyter/*.gif
 site/*
-pan3d/ui/pan3d_components/serve
+pan3d/ui/pan3d_components/module/serve
```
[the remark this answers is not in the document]
jourdain: `pan3d/ui/pan3d_components/module/serve` => `pan3d/ui/module/serve`

pan3d-js/vite.config.js
```diff
@@ -15,7 +15,7 @@ export default {
         },
       },
     },
-    outDir: "../pan3d/ui/pan3d_components/serve",
+    outDir: "../pan3d/ui/pan3d_components/module/serve",
```
jourdain, line 18: don't forget that if you remove `pan3d_components` you will have to fix many location

.gitignore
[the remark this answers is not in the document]
annehaley: I added it as a container for both the `module` and `serve` directory, but now I guess its only purpose is for the name; it specifies the purpose of the `module` directory. The name `module` is pretty vague. Maybe I could remove the extra `module` dir and `pan3d_components` could contain `serve` and `__init__.py` directly? But I'm not sure if the `module` directory is mandatory.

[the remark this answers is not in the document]
annehaley: Or any other name would suffice to indicate what this directory contains, I am open to suggestions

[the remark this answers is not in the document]
annehaley: Ok, if we cannot change the name "module", then I'd rather keep it inside a shell directory that has a more precise name.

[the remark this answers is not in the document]
jourdain: so you should put widgets.py in it for consistency. One can not work without the other.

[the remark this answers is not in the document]
annehaley: Ok, moved in [a28e308](https://github.com/Kitware/geo-trame/pull/78/commits/a28e3083212c6c2dd0556228951f85d9f62a15d9)

pan3d/ui/pan3d_components/widgets.py
```diff
@@ -0,0 +1,14 @@
+from trame_client.widgets.core import HtmlElement
```
jourdain, line 1: I think that is the wrong base class and you are missing enabling the module here.

jourdain: See https://github.com/Kitware/trame-components/blob/master/trame_components/widgets/trame.py#L20-L24

annehaley: The module is enabled in the DatasetViewer initialization. Which base class should we use instead?

jourdain: See comment above, that is what is expected.

jourdain:
```python 
from trame_client.widgets.core import AbstractElement
from . import module

__all__ = [
    "PreviewBounds",
]


class HtmlElement(AbstractElement):
    def __init__(self, _elem_name, children=None, **kwargs):
        super().__init__(_elem_name, children, **kwargs)
        if self.server:
            self.server.enable_module(module)


class PreviewBounds(HtmlElement):
    ...
```

pyproject.toml
```diff
@@ -63,7 +63,7 @@ requires = ["setuptools>=61", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools]
-packages = ["pan3d", "pan3d.ui", "pan3d.catalogs"]
+packages = ["pan3d", "pan3d.ui", "pan3d.ui.pan3d_components", "pan3d.ui.pan3d_components.module", "pan3d.catalogs"]
```
jourdain, line 66: Is just listing "pan3d" not enough?

pan3d/ui/pan3d_components/widgets.py
annehaley: Alright, I applied this suggestion in [1c4d30c](https://github.com/Kitware/geo-trame/pull/78/commits/1c4d30cbab102623c7c4be0aaf73ef4fe768da1f)

pyproject.toml
annehaley: In my experience, this results in an import error when attempting to import something in any constituent folders.

annehaley: For example, https://github.com/Kitware/geo-trame/actions/runs/9570591491/job/26385698536 failed prior to [8389496](https://github.com/Kitware/geo-trame/pull/78/commits/8389496ac09e0b4988c4d733fd6ac0c91f34d3bf)

jourdain (approved):

johnkit (approved):

PR #89, opened 2024-07-23
Adding exception handling for cftime to datetime conversion

ayenpure:
It appears that certain cftime objects do not readily support conversion to DateTime objects. This change just returns the string for the cftime object in such a failure.
This also fixes two failing examples:
`example_config_cmip` and `example_config_esgf`

### DIFF
--- a/pan3d/dataset_viewer.py
+++ b/pan3d/dataset_viewer.py
@@ -204,14 +204,16 @@ class DatasetViewer:
             catalog_id = self.state.catalog.get("id")
             search_options = pan3d_catalogs.get_search_options(catalog_id)
             self.state.available_catalogs = [
-                {
-                    **catalog,
-                    "search_terms": [
-                        {"key": k, "options": v} for k, v in search_options.items()
-                    ],
-                }
-                if catalog.get("id") == catalog_id
-                else catalog
+                (
+                    {
+                        **catalog,
+                        "search_terms": [
+                            {"key": k, "options": v} for k, v in search_options.items()
+                        ],
+                    }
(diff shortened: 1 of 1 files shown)


pan3d/dataset_viewer.py
```diff
@@ -501,6 +502,22 @@ def _dataset_changed(self) -> None:
         else:
             self.state.dataset_ready = False
 
+    def _get_datetime_label(self, dtype, v) -> str:
+        if dtype.kind in ["O", "M"]:  # is datetime
+            try:
+                return (pandas.to_datetime(v).strftime("%b %d %Y %H:%M"))
+            except Exception as e:
+                # Get around the case where certain cftime objects do not
+                # readily agree with conversion to datatime objects
```
annehaley, line 511: Very small suggestion: `datatime` -> `datetime`

johnkit:
Just curious -- have we considered using the `cftime` library to parse these strings?

ayenpure:
> strftime

ayenpure:
@johnkit I don't think we need to use any of the `cftime` parsing features for now as these objects return a string representation when asked to be case to a string using `str()` I think the logic here also works for more general cases where the dimension already is in python `datetime`  format. That's when numpy would report type as `dtype(M)`.
However I don't know how many cases we'll encounter with `dtype(O)` and what the correct course would be for representing time. That's why I left the original type check for `[O, M]` instead of just `M` so we'll still follow the same behavior for anything other than `cftime` and `datetime`

ayenpure:
@annehaley @johnkit ready to be merged?

annehaley (approved):

PR #83, opened 2024-06-28
Rename to GeoTrame and update documentation

annehaley:
This PR uses the name "GeoTrame" to refer to the `DatasetViewer` in user-facing portions of the code, including the console script to start the viewer, as well as the title of the Trame app. With this rename comes a sweeping documentation update, since all screenshots needed to be updated to show the new toolbar text. With the documentation update, I've added new portions to describe new features like cartographic mode, auto slicing, and the bounds configuration menu.

To view the docs locally, checkout this branch and run `mkdocs serve`. Then, navigate to localhost:8000 in your browser.

PR #91, opened 2024-07-25
Rename and docs update rebase

ayenpure:
This is a duplicate of MR #83 
It contains the changes for renaming `pan3D-viewer` to `geotrame`  and updating documentation wherever necessary
I created the MR from a new branch so that we do not lose Anne's branch in case I messed up the rebase.


johnkit (approved):
I know Abhi walked through everything, so let's go ahead and merge.

ayenpure:
@annehaley Can you take a look once more and approve if this looks okay?

pyproject.toml
```diff
@@ -49,25 +49,11 @@ pangeo = [
...
-
-    # esgf
-    "intake-esgf>=2024.1",
-
-    # pangeo
-    "intake==0.7.0",  # latest has plugin errors
-    "intake-xarray>=0.7",
-    "intake-esm>=2023.11",
-    "s3fs>=2024.2",
-    "gcsfs>=2024.2",
```
annehaley, line 66: I think this change is outdated, so you can take this out. I believe @johnkit had mentioned that the recursive dependency definition doesn't work in some cases, so we should use the explicit list.

PR #92, opened 2024-07-30
Janelia Catalog

annehaley:
This is a WIP for @ayenpure to work with; this branch can either be continued or abandoned. 

This is just the start of a catalog module for Janelia datasets. The required catalog functions `get_catalog`, `get_search_options`, and `search` are complete, but the `load_dataset` function is incomplete. I attempted to do some data reformatting to save the data in a format that xarray could read, but I couldn't get it to work.


ayenpure:
@annehaley Thanks for sharing :)

PR #103, opened 2024-08-15
Fixing ESGF example : Set time dimension to t instead of z

ayenpure:
This change sets the time dimension of the data to the T coordinate
  instead of the Z coordinate while the example launches.


johnkit (approved):
I see there is also a "z_scale" option under the "render" settings. We could probably delete that line too. Either way, yes, let's switch this from the half-baked cube display to 2D vs time.

PR #115, opened 2024-09-25
feat(Slice Explorer): adding inital version of the slice explorer

ayenpure:
Adding Slice Explorer
  -- the main module in the `explorers` directory
  -- a driver example using python in `examples`
  -- a jupyter notebook demo example


examples/slice_explorer.py
```diff
@@ -0,0 +1,24 @@
+from argparse import ArgumentParser, BooleanOptionalAction
+
+from pan3d import DatasetBuilder
+from pan3d import SliceExplorer
+
+def serve():
+    parser = ArgumentParser(
```
jourdain, line 7: Technically you already have a parser on `server.cli`

pan3d/__init__.py
```diff
@@ -1,15 +1,16 @@
...
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
 __version__ = "0.8.9"
 
 try:
     from .dataset_viewer import DatasetViewer
-
-    __all__ = [DatasetBuilder, DatasetViewer]
+    from .explorers.slice_explorer import SliceExplorer
```
jourdain, line 12: I don't think it is a good idea to put it at the top package especially since trame is supposed to be optional for pan3d in case you only care about the dataset_builder.

jourdain: I see the try catch now... Never mind.

pan3d/dataset_builder.py
```diff
@@ -319,6 +319,41 @@ def t_index(self, t_index: int) -> None:
                 self._viewer._time_index_changed()
                 self._viewer._mesh_changed()
 
+    @property
+    def t_num(self) -> int:
```
jourdain, line 323: t_size, t_len ?

examples/slice_explorer.py
```diff
@@ -0,0 +1,24 @@
...
+
+    parser.add_argument("--config_path")
+    parser.add_argument("--server", action=BooleanOptionalAction)
+    parser.add_argument("--debug", action=BooleanOptionalAction)
+    args = parser.parse_args()
+
+    builder = DatasetBuilder()
+    builder.import_config(args.config_path)
+
+    viewer = SliceExplorer(builder)
```
jourdain, line 20: Maybe slice explorer could work without DatasetBuilder and just create+configure it if not provided using the `--config-path` arg

pan3d/explorers/slice_explorer.py
```diff
@@ -0,0 +1,441 @@
...
+]
+
+class LookupTable:
+    Rainbow = 0
+    Inverted_Rainbow = 1
+    Greyscale = 2
+    Inverted_Greyscale = 3
+
+# Color Map Callbacks
+def use_preset(sactor : vtk.vtkActor, dactor : vtk.vtkActor, preset : str) -> None:
```
jourdain, line 22: Wondering if we could have better preset defined in a json map or something.

pan3d/explorers/slice_explorer.py
```diff
@@ -0,0 +1,441 @@
...
+        print("Setting log scale")
+        lut.SetScaleToLog10()
+    else:
+        print("Setting linear scale")
+        lut.SetScaleToLinear()
+    lut.Build()
+
+@TrameApp()
+class SliceExplorer:
+    def __init__(self, builder : DatasetBuilder, server=None):
```
jourdain, line 68: allow the builder to be none

pan3d/explorers/slice_explorer.py
```diff
@@ -0,0 +1,441 @@
...
+            float(extents[2][0] + (extents[2][1] - extents[2][0]) / 2)
+        ] 
+        self.normal = [1, 0, 0]
+        self.lengths = [
+            float(extents[0][1] - extents[0][0]),
+            float(extents[1][1] - extents[1][0]),
+            float(extents[2][1] - extents[2][0]),
+        ]
+
+        self.plane = vtk.vtkPlane()
```
jourdain, line 104: I would gather the vtk code inside `self._setup_viz()` and just call it in constructor

pan3d/explorers/slice_explorer.py
```diff
@@ -0,0 +1,441 @@
...
+        self.state.dimval = value
+        self.on_data_change()
+
+    @property
+    def view_mode(self):
+        return self.state.view_mode
+    
+    @view_mode.setter
+    def view_mode(self, mode):
+        self.state.view_mode = mode
```
jourdain, line 208: I would nest all the setters content inside `with self.state:` to allow out jupyter usage

pan3d/explorers/slice_explorer.py
```diff
@@ -0,0 +1,441 @@
...
+            float(extents[0][0] + (extents[0][1] - extents[0][0]) / 2), 
+            float(extents[1][0] + (extents[1][1] - extents[1][0]) / 2), 
+            float(extents[2][0] + (extents[2][1] - extents[2][0]) / 2)
+        ]
+
+        camera = self.renderer.GetActiveCamera()
+        pos = camera.GetPosition()
+        norm = np.linalg.norm(np.array(origin) - np.array(pos))
+        print(f"Origin and norm : {origin, norm}")
+        if self.state.view_mode == '3D':
```
jourdain, line 246: I would extract that vtk section into helper methods `self._setup_view3d()`, `self._setup_view2d(axis)`

jourdain:
```
pos = [*origin]
pos[axis] = norm
view_up = [0,0,1] if axis == 1 else [0,1,0]
camera.SetPosition(pos)
camera.SetViewUp(view_up)
```

pan3d/explorers/slice_explorer.py
```diff
@@ -0,0 +1,441 @@
...
+            float(extents[1][0] + (extents[1][1] - extents[1][0]) / 2), 
+            float(extents[2][0] + (extents[2][1] - extents[2][0]) / 2)
+        ] 
+        origin[slice_i] = dimval
+        
+        self.plane.SetOrigin(origin)
+        self.plane.SetNormal(normal)
+
+        self.cutter.SetInputData(self.t_slice)
+        self.cutter.SetCutFunction(self.plane)
```
jourdain, line 287: No needed as already set

pan3d/explorers/slice_explorer.py
```diff
@@ -0,0 +1,441 @@
...
+        
+        extents = list(self.extents.values())
+        origin = [
+            float(extents[0][0] + (extents[0][1] - extents[0][0]) / 2), 
+            float(extents[1][0] + (extents[1][1] - extents[1][0]) / 2), 
+            float(extents[2][0] + (extents[2][1] - extents[2][0]) / 2)
+        ] 
+        origin[slice_i] = dimval
+        
+        self.plane.SetOrigin(origin)
```
jourdain, line 283: do we need to update origin/normal?

pan3d/explorers/slice_explorer.py
```diff
@@ -0,0 +1,441 @@
...
+        
+    @property
+    def levcoords(self):
+        return self.builder.dataset.coords["level"].to_numpy()
+
+    @property
+    def timecoords(self):
+        return self.builder.dataset.coords["month"].to_numpy()
+    
+    def UpdateSlicer(self, dim):
```
jourdain, line 328: methods in python should be `update_slicer`

pan3d/explorers/slice_explorer.py
```diff
@@ -0,0 +1,441 @@
...
+        self.ctrl.view_update()
+
+    @change("varmin", "varmax")
+    def on_scalar_change(self, varmin, varmax, **_):
+        self.slice_mapper.SetScalarRange(float(varmin), float(varmax))
+        self.sbar_actor.SetLookupTable(self.slice_mapper.GetLookupTable())
+        self.ctrl.view_update()
+        
+    @property
+    def levcoords(self):
```
jourdain, line 321: what about `coords_level`

pan3d/explorers/slice_explorer.py
```diff
@@ -0,0 +1,441 @@
...
+        return self.builder.dataset.coords["level"].to_numpy()
+
+    @property
+    def timecoords(self):
```
jourdain, line 325: what about `coords_time`

pan3d/explorers/slice_explorer.py
```diff
@@ -0,0 +1,441 @@
...
+        ext = self.extents[dim]
+        self.state.dimval = float(ext[0] + (ext[1] - ext[0]) / 2)
+        self.state.dimmin = float(ext[0])
+        self.state.dimmax = float(ext[1])
+
+    def start(self, **kwargs):
+        self.ui.server.start(**kwargs)
+
+    @property
+    def ui(self):
```
jourdain, line 338: I would define it inside `self._build_ui()` and just store the layout to self.ui

pan3d/explorers/slice_explorer.py
```diff
@@ -0,0 +1,441 @@
...
+        if self._ui:
+            return self._ui
+
+        style = dict(density="compact", hide_details=True)
+
+        with SinglePageWithDrawerLayout(self.server, full_height=True) as layout:
```
jourdain, line 344: Does the `full_height=True` remove the slider in your view? If not we need to add a client.Style("html, body {  overflow: hidden; }")

pan3d/explorers/slice_explorer.py
ayenpure: It was one of our plans to add some Divergent color maps similar to the ones defined here
https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/ColorTransferFunction/LiteColorMaps.json

pan3d/explorers/slice_explorer.py
ayenpure: The plane's origin, and normal is what the cutter uses to get the slice. I think every time the user changes the slice value along a dimension, we need to update the `origin`, and every time the dimension is updated, the `normal` needs to change as well.

pan3d/explorers/slice_explorer.py
```diff
@@ -0,0 +1,441 @@
...
+        if self.state.view_mode == "2D":
+            self.on_view_mode_change("2D")
+            self.renderer.ResetCamera()
+
+        self.ctrl.view_update()
+
+    @change("outline", "tdata")
+    def on_rep_change(self, outline, tdata, **_):
+        if outline: 
+            self.renderer.AddActor(self.outline_actor)
```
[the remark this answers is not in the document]
ayenpure: This is a good point, I think adding or removing actors would be helpful in large datasets as we are allowing users to have the surface rendering of the original data with some value of opacity set. Adding/removing will help control the memory of rendering artifacts, thus, also helping with performance? I think this will need a performance study as I cannot make these gusses.

pan3d/explorers/slice_explorer.py
ayenpure: I had accidentally kept this around, we do not need dimension specific coords except for time.

pan3d/explorers/slice_explorer.py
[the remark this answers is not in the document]
jourdain: I'm not talking about opacity. I don't agree with your performance assumption. Technically the visibility does the same thing as add/remove but without mutating a vtkCollection.

ayenpure:
> Before merging, we need to update slice explorer to handle more general time coordinates
> 
> 1. Update the current logic from `month` to instead use the builder's time coordinate.

I've updated the code to use `builder.t` so that we do not hard code any dimension names

> 2. Need to handle more than integer time coordinates. At a min, let's get slice explorer to at least recognize/support  `datetime64[ns]`. Maybe we can use the same logic as dataset viewer.
> 

With the change above we should be able to use any date time format. I need to check this with dataset with datatime object for time coordinates. I'll test with the new files you've shared below

> For testing, I have a small data file and notebook. For some reason, github will not let me upload either of these files, nor a tgz. So I put them on data.kitware. You can get them at https://data.kitware.com/#user/55dcb06e8d777f6ddc3ff892/folder/6703f95cfb903c47575a91f9

pan3d/dataset_builder.py
```diff
@@ -327,6 +362,16 @@ def slicing(self) -> Dict[str, List]:
         integers or floats representing start value, stop value, and step.
         """
         return self._algorithm.slicing
+    
+    @property
+    def extents(self) -> map:
+        extents = {}
```
jourdain, line 368: Does not look like a vtk extents?

pan3d/explorers/slice_explorer.py
```diff
@@ -0,0 +1,495 @@
...
+        "Range" : (0.0, 1.0),
+    },
+    "Inv Greyscale" : {
+        "Hue" : (0.0, 0.666),
+        "Saturation" : (0.0, 0.0),
+        "Range" : (1.0, 0.0),
+    },
+}
+
+def use_preset(sactor: vtk.vtkActor, dactor: vtk.vtkActor, preset: str) -> None:
```
jourdain, line 45:
```python
preset = hsv_colors[preset]
hue = preset["Hue"] 
sat = preset["Saturation"]
rng = preset["Range"]

for lut in luts:
  lut.SetHueRange(*hue)
  lut.SetSaturationRange(*sat)
  lut.SetValueRange(*rng)
  lut.Build()
```

pan3d/explorers/slice_explorer.py
```diff
@@ -0,0 +1,495 @@
...
+        print("Setting linear scale")
+        lut.SetScaleToLinear()
+    lut.Build()
+
+
+@TrameApp()
+class SliceExplorer:
+    def __init__(self, builder: DatasetBuilder = None, server=None):
+        self.server = get_server(server)
+        self.builder = builder
```
jourdain, line 73: what if builder is none? Can you create it here?

pan3d/explorers/slice_explorer.py
jourdain: @ayenpure still not fixed

pan3d/dataset_builder.py
ayenpure: I made so intentionally as our dimensions are named and the users might be interested in getting extents by the keys for the respective dimension. Also in case of, also this allows us to only return results for as many dimensions as the data has instead of always returning 6 doubles. 
But I can make it `VTK` style so that VTK users find it familiar. Thoughts?

pan3d/dataset_builder.py
ayenpure: I settled for `t_size`, seems easier to identify

pan3d/explorers/slice_explorer.py
ayenpure: Added the `LiteColorMaps`  from VTK.js

pan3d/explorers/slice_explorer.py
ayenpure: Do you mean just like 
```
  @view_mode.setter
    def view_mode(self, mode):
      with self.state:
        self.state.view_mode = mode
        self.on_view_mode_change(mode)
```

examples/slice_explorer.py
ayenpure: This example exists outside of `Trame` context and the server is built inside the `SliceExplorer` instance.

pan3d/explorers/slice_explorer.py
ayenpure: Creating is the correct path is provided in the arguments to the app.

pan3d/explorers/slice_explorer.py
ayenpure: I tried setting the property on the client, but still see the scroll bar on the right.

pan3d/explorers/slice_explorer.py
jourdain: yes

pan3d/explorers/slice_explorer.py
jourdain: ok then add the `client.Style()` unless your layout leak outside the main content, in which case, you should fix that part first.

pan3d/dataset_builder.py
jourdain: That is ok, I was just making sure it was intentional. 

You should add a comment on the python property to explain what you get.

pan3d/explorers/slice_explorer.py
johnkit: @ayenpure Please make a formal issue for this so that we can move ahead on this PR.

Side note: this scrollbar does NOT show up in my firefox browser. (Some other unexpected things occur, but I'll detail later if needed).

ayenpure:
@jourdain Any idea why the CI could be failing? 
I am at a loss as `black` locally does not report any changes for me but is failing `pre-commit` and for the MacOS tests something is weird as they pass for Linux and Windows

jourdain:
Install pre-commits - See https://github.com/Kitware/pan3d/blob/main/CONTRIBUTING.rst

Then if you run black by hand, you need to have the same version as the one defined [here](https://github.com/Kitware/pan3d/blob/main/.pre-commit-config.yaml#L3)

ayenpure:
> Install pre-commits - See https://github.com/Kitware/pan3d/blob/main/CONTRIBUTING.rst
> 
> Then if you run black by hand, you need to have the same version as the one defined [here](https://github.com/Kitware/pan3d/blob/main/.pre-commit-config.yaml#L3)

Thanks for the helpful tip, that got the formatting problems resolved.
Any idea on how to tackle the MacOS failures? Locally they seem to work without problem

jourdain:
Maybe not related, but one issue that you have is your JSON file for the preset is not getting bundled in your Python package. 
You need to edit https://github.com/Kitware/pan3d/blob/main/pyproject.toml#L80-L83

Also a possibility for mac is a dependency not being available?

pan3d/explorers/slice_explorer.py
ayenpure: Done!

johnkit (approved):
Excellent work, Abhi. thanks.

PR #137, opened 2024-11-13
Updating Slice Explorer to use new xarray APIs

ayenpure:
Updating Slice Explorer to use new xarray APIs


examples/example_slice_explorer.py
```diff
@@ -1,7 +1,6 @@
 from argparse import ArgumentParser, BooleanOptionalAction
 
-from pan3d import DatasetBuilder
-from pan3d import SliceExplorer
+from pan3d.explorers.slicer import XArraySlicer
```
jourdain, line 3: What is the point of that file? You can run it from the cli by doing 

`xr-slicer` or `python -m  pan3d.explorers.slicer`

jourdain:
The rest is good to go

PR #141, opened 2024-11-21
Update examples

johnkit:
Update pangeo and esgf examples to use latest config syntax and add some markdown cells

Pangeo Example

ESGF Example

### DIFF
--- a/pan3d/viewers/preview.py
+++ b/pan3d/viewers/preview.py
@@ -23,9 +23,11 @@ from trame.app import get_server, asynchronous
 
 from trame.ui.vuetify3 import VAppLayout
 from trame.widgets import vuetify3 as v3
+from trame.widgets import vtk as vtkw, vtklocal as vtkwasm
 
 from pan3d.xarray.algorithm import vtkXArrayRectilinearSource
 
+from pan3d.utils.constants import has_gpu
 from pan3d.utils.convert import update_camera, to_image, to_float
 from pan3d.utils.presets import apply_preset
 
@@ -89,6 +91,14 @@ class XArrayViewer:
             self.local_rendering = "wasm"
         if args.vtkjs:
             self.local_rendering = "vtkjs"
+
+        # If no GPU, use local rendering
+        if self.local_rendering is None and not has_gpu():
+            self.local_rendering = "wasm"
+
+        # Initialize conditional widgets
+        vtkw.initialize(self.server)
+        vtkwasm.initialize(self.server)
 
         # Process CLI
         self.ctrl.on_server_ready.add(self._process_cli)
--- a/examples/example_config_esgf.json
+++ b/examples/example_config_esgf.json
@@ -1,12 +1,59 @@
 {
   "data_origin": {
     "source": "esgf",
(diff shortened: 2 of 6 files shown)


jourdain:
I don't think you want to trigger a new PyPI release of pan3d for examples/docs update. You should use a commit message like `docs(examples): update pangeo example to enable viz`

johnkit:
Hacked in a commit to detect binder and hard-code local/vtkjs rendering. Good enough enable notebooks to work in binder. I didn't use the `has_gpu_rendering()` method in `utils`, which is outdated. I'll look into VTK for the general approach.

pan3d/viewers/preview.py
```diff
@@ -85,9 +86,11 @@ def __init__(self, server=None, local_rendering=None):
         )
         args, _ = self.server.cli.parse_known_args()
         self.local_rendering = local_rendering
-        if args.wasm:
+        if os.environ.get("BINDER_REQUEST") is not None:
+            self.local_rendering = "vtkjs"
```
jourdain, line 90: why do you want to use vtkjs instead of wasm for binder?

johnkit: On my first test w/WASM (before I added this commit, I used the `local_rendering` argument), the renderwindow didn't display. Definitely possible I overlooked something, so I'll change to wasm and retest.

johnkit: So "vtkjs" is ok but "wasm" isn't. It looks like part or all of `trame_vtk_local` isn't getting installed, or is installed in the wrong place. I'll paste the js console messages below, but basically Jupyter aborts because it does not find `__trame_vtklocal/js/style.css`. 

```
index-de5ee105.js:1 
        
        
       GET https://hub.binder.opensci.2i2c.cloud/user/kitware-pan3d-irtfxelw/trame-jupyter-server/servers/esgf-viewer/__trame_vtklocal/js/style.css net::ERR_ABORTED 404 (Not Found)
(anonymous) @ index-de5ee105.js:1
mo @ index-de5ee105.js:1
Ce @ index-de5ee105.js:1
So @ index-de5ee105.js:1
Fs @ index-de5ee105.js:23
await in Fs
(anonymous) @ index-de5ee105.js:23
index-de5ee105.js:23 Uncaught (in promise) Event {isTrusted: true, type: 'error', target: link, currentTarget: null, eventPhase: 0, …}
Fs @ index-de5ee105.js:23
await in Fs
(anonymous) @ index-de5ee105.js:23
```

jourdain: Do we want `os.environ.get("BINDER_REQUEST")` to superseed user input for local rendering request?

jourdain: I think I know why wasm could be missing in the install. 

Let's add `vtklocal.initialize(self.server)` somewhere since the widget get conditionally instantiated.

jourdain:
I think the 2 pending changes are:
- Should the binder selection to local rendering be more of a fallback than an actual override.
- Fix wasm to properly expose itself to overcome the conditional instantiation. (but I'm still ensure why it is working on my end with binder)

jourdain (approved):

Issue #142, opened 2024-12-03
Improvements to the globe explorer (round 2)


aashish24:

jourdain:
- [ ] Make sure the scrollbar on linux don't show up
- [ ] Title should be `Globe Explorer` instead of `XArray Viewer`
- [ ] Opacity
- [ ] Scaling might not be needed
- [ ] Multi-variables?

PR #146, opened 2024-12-16
Adding updates to globe explorer

ayenpure:
Adding partial list of features defined in https://github.com/Kitware/pan3d/issues/142


pan3d/explorers/globe.py
```diff
@@ -300,34 +297,33 @@ def _on_color_by(self, color_by, **__):
             self.state.color_min = 0
             self.state.color_max = 1
 
-    @change("color_preset", "color_min", "color_max", "nan_color")
+    @change("color_preset", "color_min", "color_max", "nan_color", "opacity")
```
jourdain, line 300: opacity could have been in its own callback as there is no relationship with color and range.

pan3d/ui/globe.py
```diff
@@ -0,0 +1,883 @@
...
+                                        color=(
+                                            "color[3] ? `rgb(${color[0] * 255}, ${color[1] * 255}, ${color[2] * 255})` : undefined",
+                                        ),
+                                        flat=True,
+                                        icon=True,
+                                        ripple=False,
+                                        size="small",
+                                        click="toggle",
+                                    )
+            v3.VSelect(
```
jourdain, line 382: no tooltip?

pan3d/utils/globe.py
```diff
@@ -64,6 +65,32 @@ def get_globe_texture():
     texture.InterpolateOn()
 
     return texture
+"""
+
+
+def get_globe_textures():
```
jourdain, line 71: should you cache those textures that never change instead of creating them every time you call that method?

pan3d/ui/globe.py
```diff
@@ -0,0 +1,883 @@
...
+                disabled=("!data_origin_id?.length || !can_load",),
+                color=("can_load ? 'primary': undefined",),
+                click=(
+                    load_dataset,
+                    "[data_origin_source, data_origin_id, data_origin_order]",
+                ),
+            )
+
+
+class DataInformation(CollapsableSection):
```
jourdain, line 151: Is that the same as the preview? Can you import and use the one from preview instead?

pan3d/ui/globe.py
```diff
@@ -0,0 +1,883 @@
...
+                    clearable=True,
+                    hide_details=True,
+                    density="compact",
+                    flat=True,
+                    variant="solo",
+                    max_width=200,
+                )
+
+
+class DataOrigin(CollapsableSection):
```
jourdain, line 82: Is that the same as the preview? Can you import and use the one from preview instead?

pan3d/ui/globe.py
```diff
@@ -0,0 +1,883 @@
...
+from pan3d import catalogs as pan3d_catalogs
+from pan3d.utils.constants import XYZ, SLICE_VARS
+from pan3d.utils.convert import max_str_length
+from pan3d.utils.presets import PRESETS
+
+from pan3d.ui.css import base, preview
+from pan3d.ui.collapsible import CollapsableSection
+
+
+class SummaryToolbar(v3.VCard):
```
jourdain, line 16: Is that the same as the preview? Can you import and use the one from preview instead?

pan3d/utils/globe.py
ayenpure: They are cached, the call to this method only happens once in the `Globe Explorer` so instead of caching it in the utils class they are used as a class member in the explorer. This method won't be called every time the user changes the texture. Does that still not feel okay?

pan3d/explorers/globe.py
```diff
@@ -307,7 +305,14 @@ def _on_color_by(self, color_by, **__):
...
     @change("color_preset", "color_min", "color_max", "nan_color")
     def _on_color_preset(
-        self, nan_color, nan_colors, color_preset, color_min, color_max, **_
+        self,
+        nan_color,
+        nan_colors,
+        color_preset,
+        color_min,
+        color_max,
+        opacity,
```
jourdain, line 314: opacity can be removed as it is not used in that method

PR #152, opened 2024-12-19
Viewers xarray accessor

jourdain:


jourdain:
FYI @ayenpure 

@aashish24 @johnkit : This is to uniform the API across viewer/explorers. I'm still missing the globe, but there is more work to do there. Will try to tackle that after my meetings.

PR #153, opened 2025-01-06
feat(contour explorer): Adding general countour explorer

ayenpure:


pan3d/explorers/contour.py
```diff
@@ -76,19 +75,48 @@ def __init__(self, server=None, local_rendering=None):
...
+        if xarray is not None:
+            self.source = vtkXArrayRectilinearSource(input=xarray)
+        elif source is not None:
+            self.source = source
+        elif import_state is not None:
+            self.source = vtkXArrayRectilinearSource()
+            config = json.loads(import_state.read_text())
+            self.source.load(config)
+            try:
+                field = config["preview"]["color_by"]
```
jourdain, line 108:
```python
field = config.get("preview", {}).get("color_by")
```

jourdain: Should we have another name than "preview"?

pan3d/xarray/accessor.py
```diff
@@ -86,6 +86,19 @@ def globe(self):
 
         return self._viewer_globe
 
+    @property
+    def contour(self):
+        from pan3d.explorers.contour import ContourExplorer
+
+        if self._viewer_contour is None:
```
jourdain, line 93: missing `self._viewer_contour` in constructor.

pan3d/explorers/contour.py
jourdain: I don't see export state... It seems that the import/export is half backed.

ayenpure: I think as it stands import/export state only happens through the main preview app.
None of our explorers export a state, but they are able to import a state. We could very-well add support to export state to them as well. I can work in it in the next milestone and it should not take too long.
Although, yes `preview` should be renamed to something more generic -- `rendering`?

jourdain (approved):

pan3d/explorers/contour.py
jourdain: The globe explorer properly handle import/export.

Each explorer has its own namespace. "preview" is for the viewer, which is fine, unless we assume that piece of namespace get shared with all the rest of explorers.

jourdain: If we assume slice/contour can only import state, that could be fine too... We just need to know why we made that choice.

jourdain:
Do you want me to merge and just follow up with some more improvement in other PRs?

pan3d/explorers/contour.py
ayenpure: The Globe Explorer, for the most part, was derived directly from the Preview app, so it got whatever came along with it. However, the export state functionality could be broken because I have not updated the stored essential variables in the state. The state export was omitted for the others to keep the UI and development simple and fast.
A standard UI component for import-export can make things easier by adding callbacks from the explorers.

jourdain:
I did "rebase + merge" to see if the CI properly generate a new version

jourdain:
that properly trigger a release...

PR #154, opened 2025-01-06
fix (explorer consistency): Making explorer name and constructor arguments consistent

ayenpure:
Making the explorers more consistent in naming and behavior

### DIFF
--- a/pan3d/explorers/globe.py
+++ b/pan3d/explorers/globe.py
@@ -42,7 +42,7 @@ vtkObject.GlobalWarningDisplayOff()
 
 
 @TrameApp()
-class GlobeViewer:
+class GlobeExplorer:
     """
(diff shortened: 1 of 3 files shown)


jourdain (changes requested):
Current code is good but you are forgetting to update that section

https://github.com/Kitware/pan3d/blob/main/pyproject.toml#L69-L74

jourdain (approved):
Actually the file name did not change and since the `main` method is what is getting executed, you are fine.

jourdain:
I'm not sure why a new released was not triggered.

Just some possible guess:
1. I merged instead of rebase which changed the commit and no version change was detected
2. The commit message was not properly formatted `fix (explorer consistency): ...` instead of `fix(explorers): ...`

PR #156, opened 2025-01-07
feat(reader): use vtk implementation if available

jourdain:


pan3d/xarray/vtk.py
```diff
@@ -0,0 +1,536 @@
...
+            if self._pipeline is not None:
+                output = self._pipeline(mesh)
+            else:
+                output = mesh
+
+            # set it as output
+            print("output => ", output.GetClassName())
+            filter_output = VTK_DATASETS[output.GetClassName()].GetData(outInfo)
+            print(f"{filter_output=}")  # <= it is None...
+            filter_output.ShallowCopy(output)
```
jourdain, line 531: @berkgeveci I can't seem to set the filter output properly. The main issue is that I don't know the type of the output before hand. It can be a `vtkImageData` or a `vtkUnstructuredGrid` or else.

When I was setting the output type to vtkDataObject, when I was doing a ShallowCopy, my vtkImageData got converted to a vtkDataObject and lost of what I care from it... 

Do you know a proper way to do that?

pan3d/ui/preview.py
```diff
@@ -776,6 +776,10 @@ def on_change(self, slice_t, **_):
 
         self.source.slices = slices
         ds = self.source()
+
+        if ds.GetClassName() == "vtkDataObject":
+            return  # no mesh produced yet
```
jourdain, line 781: That can probably go away once the reader output get properly generated.

pan3d/viewers/preview.py
```diff
@@ -267,6 +272,9 @@ def _on_color_by(self, color_by, **__):
             return
 
         ds = self.source()
+        print("=" * 60)
+        print(ds)
+        print("=" * 60)
```
jourdain, line 277: remove debug print but help validate what the source generate...

pan3d/viewers/preview.py
```diff
@@ -267,6 +272,9 @@ def _on_color_by(self, color_by, **__):
...
         if color_by in ds.point_data.keys():
```
jourdain, line 278: The data is located on cells with the VTK reader... Need to be flexible!

pan3d/xarray/vtk.py
jourdain: Original reader that I'm wrapping https://github.com/Kitware/VTK/blob/master/IO/NetCDF/vtkNetCDFReader.cxx#L166-L180

berkgeveci: You need to implement RequestDataObject() and create an output of the same type as the output of the underlying reader. In that method, first call UpdateDataObject() on the _reader. Then, get its output and make a copy of it using NewInstance(). Once you do that, this code should work properly.

jourdain:
@aashish24 @ayenpure we should revisit it to see what could be integrated with the latest code base at a reasonable cost.

One of the usage is to be able the ROMS netcdf file. For PNNL...

PR #157, opened 2025-01-10
Fix removed dimensions property for VTK structured grid

danlipsa:


jourdain:
The code looks good, but I think I still want to see the method available on the Python side.

Side note, the commit message needs to respect semantic-release so the CI can update the changelog and trigger a release if need be.

danlipsa:
> The code looks good, but I think I still want to see the method available on the Python side.

Do you mean to add back the .dimensions property that was deprecated and removed?


> Side note, the commit message needs to respect semantic-release so the CI can update the changelog and trigger a release if need be.
Sounds good, I'll change that.

jourdain:
Yes on the Python override, not on the C++.

jourdain:
Do you understand why it was removed on the C++ side?

danlipsa:
> Do you understand why it was removed on the C++ side?

No, I don't. I didn't look at that code much.

jourdain:
I didn't look at the code either, but based on the method profile it is obvious.

Issue #158, opened 2025-01-14
Deliver Analytics Explorer MVP 


aashish24:
- [x] Infovis plots on the right side in a drawer
- [x] All the options on the top of the drawer
- [x] Options to group by for climate averages
- [x] Conda installable

aashish24:
Moving to the Spring 09

aashish24:
This has been completed. There is an issue of performance, but that will be a separate issue.

PR #162, opened 2025-02-02
Analytics explorer

ayenpure:
A work in progress PR for the analytics explorer
ToDo:
- [ ] re-add caching of figure/date for when data does not change
- [ ] performance in zonal and global averages with full temporal resolution
- [x] customize Pan3DView to contain additional button to toggle plot drawer
- [x] Add warnings for when `xcdat` not available

### DIFF
--- a/pan3d/explorers/analytics.py
+++ b/pan3d/explorers/analytics.py
@@ -56,10 +56,8 @@ class Pan3dAnalyticsView(Pan3DView):
 
 @TrameApp()
 class AnalyticsExplorer:
-    """Create a Trame GUI for a Pan3D XArray Viewer"""
-
     def __init__(self, xarray=None, server=None, local_rendering=None):
-        """Create an instance of the XArrayViewer class.
+        """Create an instance of the AnalyticsExplorer class.
 
         Parameters:
             server (str/server): Trame server name or instance.
@@ -200,9 +198,7 @@ class AnalyticsExplorer:
     # -------------------------------------------------------------------------
 
     def _build_ui(self, **kwargs):
-        print("Control : ", self.state.control_expended)
-
-        self.state.trame__title = "XArray Viewer"
+        self.state.trame__title = "Analytics Explorer"
 
         with VAppLayout(self.server, fill_height=True) as layout:
             self.ui = layout
--- a/pan3d/explorers/analytics.py
+++ b/pan3d/explorers/analytics.py
@@ -23,7 +23,7 @@ from trame.decorators import TrameApp, change
 from trame.app import get_server, asynchronous
 
 from trame.ui.vuetify3 import VAppLayout
-from trame.widgets import vuetify3 as v3
+from trame.widgets import vuetify3 as v3, html
 
 from pan3d.xarray.algorithm import vtkXArrayRectilinearSource
 
@@ -34,6 +34,24 @@ from pan3d.utils.presets import set_preset
 from pan3d.ui.vtk_view import Pan3DView, Pan3DScalarBar
 from pan3d.ui.preview import SummaryToolbar, ControlPanel
 from pan3d.ui.analytics import Plotting
+
+"""
+@TrameApp()
+class Pan3dAnalyticsView(Pan3DView):
+    def __init__(self, render_window, **kwargs):
+        super().__init__(render_window=render_window, **kwargs)
+        with self.toolbar:
+            v3.VDivider(classes="my-1")
+            with v3.VTooltip(text="Look toward at an angle"):
+                with html.Template(v_slot_activator="{ props }"):
+                    v3.VBtn(
+                        v_bind="props",
+                        flat=True,
+                        density="compact",
+                        icon="mdi-axis-arrow",
+                        click=(self.reset_camera_to_axis, "[[1,1,1]]"),
+                    )
+"""
 
 
 @TrameApp()
@@ -115,6 +133,8 @@ class AnalyticsExplorer:
         self.ui = None
         self._setup_vtk()
         self._build_ui()
+        print("forcing update figure", self.state.color_by)
+        self.plotting.on_change_active_plot()
 
     # -------------------------------------------------------------------------
     # VTK Setup
@@ -186,21 +206,6 @@ class AnalyticsExplorer:
 
         with VAppLayout(self.server, fill_height=True) as layout:
             self.ui = layout
-
-            # 3D view
-            Pan3DView(
-                self.render_window,
-                local_rendering=self.local_rendering,
-                widgets=[self.widget],
-            )
-
-            # Scalar bar
-            Pan3DScalarBar(
-                v_show="!control_expended",
-                v_if="color_by",
-                img_src="preset_img",
-            )
-
             # Save dialog
             with v3.VDialog(v_model=("show_save_dialog", False)):
                 with v3.VCard(classes="mx-auto w-50"):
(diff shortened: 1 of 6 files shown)


pan3d/explorers/analytics.py
```diff
@@ -0,0 +1,576 @@
...
+                    )
+"""
+
+
+@TrameApp()
+class AnalyticsExplorer:
+    """Create a Trame GUI for a Pan3D XArray Viewer"""
+
+    def __init__(self, xarray=None, server=None, local_rendering=None):
+        """Create an instance of the XArrayViewer class.
```
jourdain, line 62: wrong class name in comment

pan3d/explorers/analytics.py
```diff
@@ -0,0 +1,576 @@
...
+            - `--wasm`: Use WASM for local rendering
+            - `--vtkjs`: Use vtk.js for local rendering
+        """
+        self.xarray = xarray
+        self.server = get_server(server, client_type="vue3")
+        if self.server.hot_reload:
+            self.ctrl.on_server_reload.add(self._build_ui)
+
+        # cli
+        self.server.cli.add_argument(
```
jourdain, line 84: we've never done that in the previous apps, but it might be good for the CLI to create a group for those add-on args so they don't get bundled with the rest of the trame/wslink ones. The idea, will be that when we run with `--help` we can clearly identify the ones that are for that explorer/app.

pan3d/explorers/analytics.py
```diff
@@ -0,0 +1,576 @@
...
+            [0, 0, 1, 1],
+            [0.9, 0.9, 0.9, 0],
+        ]
+        self.state.nan_color = 2
+
+        self.ui = None
+        self._setup_vtk()
+        self._build_ui()
+        print("forcing update figure", self.state.color_by)
+        self.plotting.on_change_active_plot()
```
jourdain, line 137: This should automatically be called right before on_server_ready as there is a @state.change attached to it. You should not need to make that call here.

pan3d/explorers/analytics.py
```diff
@@ -0,0 +1,576 @@
...
+    @property
+    def state(self):
+        """Returns the current the trame server state."""
+        return self.server.state
+
+    @property
+    def ctrl(self):
+        """Returns the Controller for the trame server."""
+        return self.server.controller
+
```
jourdain, line 197: We may want to add to that list of "trame" methods [those ones](https://github.com/Kitware/pan3d/blob/main/pan3d/viewers/preview.py#L512-L517).

pan3d/explorers/analytics.py
jourdain: We could create a pan3d abstract class for an "app". That way we can get the boiler plate code done for all.

pan3d/explorers/analytics.py
```diff
@@ -0,0 +1,576 @@
...
+    # -------------------------------------------------------------------------
+    # UI
+    # -------------------------------------------------------------------------
+
+    def _build_ui(self, **kwargs):
+        print("Control : ", self.state.control_expended)
+
+        self.state.trame__title = "XArray Viewer"
```
jourdain, line 205: Do you want to use your own name?

pan3d/ui/analytics.py
```diff
@@ -0,0 +1,390 @@
+from trame.decorators import TrameApp, change
+from trame.widgets import vuetify3 as v3, plotly, html
+
+import hashlib
+import xcdat  # noqa
```
jourdain, line 5: Why do you need `xcdat` here?

pan3d/ui/css/preview.css
```diff
@@ -22,6 +22,16 @@ html {
     z-index: 2;
 }
 
+.controllerbr {
```
jourdain, line 25: is `controllerbr` a good name?

pan3d/ui/vtk_view.py
```diff
@@ -233,6 +233,9 @@ def _on_view_type_change(self, view_3d, **_):
             self.ctrl.view_reset_camera()
 
 
+print(type(Pan3DView))
```
jourdain, line 236: remove debug?

ayenpure:
This issue addresses 
https://github.com/Kitware/pan3d/issues/158

pan3d/explorers/analytics.py
```diff
@@ -0,0 +1,565 @@
...
+    # -----------------------------------------------------
+    # Public API
+    # -----------------------------------------------------
+
+    def export_state(self):
+        """Return a json dump of the reader and viewer state"""
+        camera = self.renderer.active_camera
+        state_to_export = {
+            **self.source.state,
+            "preview": {
```
jourdain, line 475: should we have our own key instead of preview?

pan3d/ui/analytics.py
```diff
@@ -0,0 +1,405 @@
...
+
+try:
+    import xcdat  # noqa
+except ModuleNotFoundError as e:
+    print(
+        f"""
+        Error occurred while importing xcdat required for Analytics Explorer: {e}
+        Please make sure xcdat is installed properly using conda -- `conda install -c conda-forge xcdat`
+        You may want to create a new conda environment an reinstall pan3d within along with xcdat.
+        """
```
jourdain, line 14: is `xcdat` available on PyPI?

pan3d/ui/analytics.py
```diff
@@ -0,0 +1,405 @@
...
+                self.ctrl.figure_update = figure.update
+
+    def expose_plot_specific_config(self):
+        """
+        Toggle visibility of components controlling plots based on chosen type
+        """
+        state = self.state
+        plot_type = state.active_plot
+        if plot_type == plot_options.get(PlotTypes.ZONAL):
+            state.figure_height = 50
```
jourdain, line 158: You could have some static setup in a map of PlotTypes so you could do something like

```python
config = PLOT_CONFIGS.get(plot_type, {})
state.update(config)
```

pan3d/ui/analytics.py
ayenpure: It is not, it's only available through conda-forge at this point.

ayenpure:
@jourdain Please review it again; most suggested changes disappeared after refactoring.

pan3d/explorers/analytics.py
```diff
@@ -0,0 +1,310 @@
...
+        self.widget.SetInteractor(self.interactor)
+        self.widget.SetViewport(0.85, 0, 1, 0.15)
+        self.widget.EnabledOn()
+        self.widget.InteractiveOff()
+
+    # -------------------------------------------------------------------------
+    # Trame API
+    # -------------------------------------------------------------------------
+
+    def start(self, **kwargs):
```
jourdain, line 100: Since you are an explorer, should those method be removed as they are part of our parent class?

pan3d/explorers/analytics.py
```diff
@@ -0,0 +1,310 @@
...
+        """Initialize the UI and start the server for XArray Viewer."""
+        self.ui.server.start(**kwargs)
+
+    @property
+    async def ready(self):
+        """Start and wait for the XArray Viewer corroutine to be ready."""
+        await self.ui.ready
+
+    @property
+    def state(self):
```
jourdain, line 110: Since you are an explorer, should those method be removed as they are part of our parent class?

pan3d/explorers/analytics.py
```diff
@@ -0,0 +1,310 @@
...
+        """Returns the current the trame server state."""
+        return self.server.state
+
+    @property
+    def ctrl(self):
```
jourdain, line 115: Since you are an explorer, should those method be removed as they are part of our parent class?

pan3d/ui/analytics.py
jourdain: crap, to bad...

pan3d/ui/analytics.py
```diff
@@ -0,0 +1,415 @@
...
+        self.spatial_cache = {}
+        self.temporal_cache = {}
+        self.zonal_cache = {}
+
+        # State variables controlling the UI for various types of plots
+        self.state.figure_height = 50
+        self.state.show_group_by = False
+        self.state.show_temporal_slider = False
+        self.state.show_zonal_axis = True
+        self.state.show_update_button = False
```
jourdain, line 94: you should use `self.state.setdefault(name, value)` instead

jourdain:
Looks good except maybe with my last comments

Issue #177, opened 2025-04-03
Slice explorer after refactor is broken


jourdain:
No color in its slicing and the extra toolbar look weird and doesn't seem to work.
Also I'm not sure it is because of the refactor, but the scrolling of the controller is not working anymore.

<img width="1155" alt="Image" src="https://github.com/user-attachments/assets/371fa09c-952c-4655-a594-c0ffb99299da" />
<img width="1155" alt="Image" src="https://github.com/user-attachments/assets/1f433ba3-528a-4543-9458-3c3206c057ad" />

ayenpure:
@jourdain I'll check what's wrong

PR #178, opened 2025-04-07
fix(slice explorer update): Slice Explorer won't load necessary data arrays

ayenpure:
This fix addresses https://github.com/Kitware/pan3d/issues/177


ayenpure:
@jourdain This fixes the data issue. As for the UI, the problems are with the custom CSS classes not being honored. Perhaps we can debug together.

PR #193, opened 2025-05-21
fix(explorer data initialization): Assign xr and vtk data in super class

ayenpure:
Adding a consistent way for all explorers to initialize from VTK Algorithm for xarray


src/pan3d/utils/common.py
```diff
@@ -83,6 +84,17 @@ def __init__(self, xarray=None, source=None, server=None, local_rendering=None):
...
+        self.xarray = None
+        self.source = vtkXArrayRectilinearSource()
+
+        if source is not None:
+            self.source = source
+            self.xarray = source.input
+        elif xarray is not None:
+            self.source = vtkXArrayRectilinearSource(input=xarray)
+            self.xarray = xarray
+
```
jourdain, line 97: that initialization seems strange.

src/pan3d/utils/common.py
```diff
@@ -826,7 +838,7 @@ def _on_array_selection(self, data_arrays, **_):
             return
 
         self.state.dirty_data = True
-        if len(data_arrays) == 1:
+        if len(data_arrays) > 0:
```
jourdain, line 841: do you always want to reset it?

jourdain: I think you may want to rework the condition logic to truly match what you aim to do.

jourdain:
```python
if len(data_arrays) == 0:
   self.state.color_by = None
elif self.state.color_by == None or self.state.color_by not in data_arrays:
       self.state.color_by = data_arrays[0]
```

src/pan3d/utils/common.py
jourdain: `self.source = None`

src/pan3d/utils/common.py
```diff
@@ -83,6 +84,17 @@ def __init__(self, xarray=None, source=None, server=None, local_rendering=None):
 
         self.ui = None
 
+        # Initialize source
+        self.xarray = None
+        self.source = vtkXArrayRectilinearSource()
```
jourdain, line 89: should be None

ayenpure: should we remove `self.xarray` altoghther?
i.e. make the explorer only rely on the source (vtkAlgorithm)

jourdain: no, having the xarray is useful

Issue #186, opened 2025-05-06
Modularize widget components for building explorers


jourdain:

jourdain:
- [ ] 3D view 
- [ ] 3D view toolbar
- [ ] Controller container
- [ ] Color selector
- [ ] range selector
- [ ] scalar bar
- [ ] summary toolbar

PR #194, opened 2025-05-23
feat (modularize): First steps to modularizing Pan3D components

ayenpure:
Addressing issue https://github.com/Kitware/pan3d/issues/186

  -- modulerized Pan3D color picker and scalar bar


src/pan3d/explorers/analytics.py
```diff
@@ -16,7 +16,7 @@
 
 from pan3d.ui.analytics import Plotting
 from pan3d.ui.preview import RenderingSettings
-from pan3d.ui.vtk_view import Pan3DScalarBar, Pan3DView
+from pan3d.ui.vtk_view import Pan3DView, ScalarBar
```
jourdain, line 19: import seems wrong

src/pan3d/explorers/globe.py
```diff
@@ -23,7 +23,7 @@
 
 from pan3d.filters.globe import ProjectToSphere
 from pan3d.ui.globe import GlobeRenderingSettings
-from pan3d.ui.vtk_view import Pan3DScalarBar, Pan3DView
+from pan3d.ui.vtk_view import Pan3DView, ScalarBar
```
jourdain, line 26: import seems wrong

src/pan3d/explorers/slicer.py
```diff
@@ -26,7 +26,7 @@
 )
 
 from pan3d.ui.slicer import SliceRenderingSettings
-from pan3d.ui.vtk_view import Pan3DScalarBar, Pan3DView
+from pan3d.ui.vtk_view import Pan3DView, ScalarBar
```
jourdain, line 29: import seems wrong

src/pan3d/viewers/preview.py
```diff
@@ -182,20 +183,21 @@ def _on_color_by(self, color_by, **__):
             self.state.color_min = 0
             self.state.color_max = 1
 
-    @change("color_preset", "color_min", "color_max", "nan_color")
-    def _on_color_preset(
+    @change("color_preset")
```
jourdain, line 186: should that be managed by the ColorBy component?

src/pan3d/widgets/color.py
```diff
@@ -0,0 +1,207 @@
...
+                                        click="toggle",
+                                    )
+
+    def reset_color_range(self):
+        """Reset the color range to the min and max values of the selected data array."""
+        state = self.state
+        color_by = state[self._color_by]
+        ds = self._source()
+
+        if color_by in ds.point_data.keys():  # vtk is missing in iter
```
jourdain, line 151: point_data/cell_data should be a parameter/option

src/pan3d/widgets/color.py
```diff
@@ -0,0 +1,207 @@
...
+                        density="compact",
+                        flat=True,
+                        variant="outlined",
+                        classes="mx-2",
+                        click=self.reset_color_range,
+                    )
+            # v3.VDivider()
+            with html.Div(classes="mx-2"):
+                html.Img(
+                    src=("preset_img", None),
```
jourdain, line 92: I don't see that handled within the component

jourdain:
I don't see much isolation from what you are doing... You are still in the Widget mode more than the Component mode.

src/pan3d/explorers/contour.py
```diff
@@ -266,20 +269,13 @@ def _on_update_data(self, color_by, time_idx, **_):
...
     def _on_update_color_range(
         self, nb_contours, color_min, color_max, color_preset, **_
     ):
-        if self.last_preset != color_preset:
-            self.last_preset = color_preset
-            set_preset(self.lut, color_preset)
-            self.state.preset_img = to_image(self.lut, 255)
-
-        self.mapper.SetScalarRange(color_min, color_max)
         self.bands.GenerateValues(nb_contours, [color_min, color_max])
```
jourdain, line 279: that won't always work. You are changing the behavior in introducing bugs here.

src/pan3d/ui/contour.py
```diff
@@ -133,25 +128,25 @@ def __init__(self, source, update_rendering):
             )
 
     def update_from_source(self, source=None):
+        state = self.state
+        source = source or self._retrieve_source()
         if source is None:
-            source = self.source
+            return
 
         with self.state:
```
jourdain, line 136: If you write `with self.state as state:` you can remove line 131.

You can also rename it to something smaller if that helps (s).

src/pan3d/explorers/analytics.py
```diff
@@ -171,10 +168,9 @@ def _build_ui(self, **kwargs):
                         )
 
                         # Scalar bar
-                        Pan3DScalarBar(
+                        self.scalar_bar = ScalarBar(
```
jourdain, line 171: You can use `ctx_name="scalar_bar"` to keep the instance around and accessible.

src/pan3d/explorers/analytics.py
```diff
@@ -210,48 +206,42 @@ def _build_ui(self, **kwargs):
                         source=self.source, toggle="chart_expanded"
                     )
 
+        self.ctrl.source_update_rendering_panel = self.rendering.update_from_source
```
jourdain, line 209: Attaching method controller is usually better to keep them close to the widget definition, so when refactoring you have no risk of doing half of the work.

src/pan3d/explorers/analytics.py
```diff
@@ -210,48 +206,42 @@ def _build_ui(self, **kwargs):
...
+            )
             self.mapper.SelectColorArray(color_by)
             self.mapper.SetScalarModeToUsePointFieldData()
             self.mapper.InterpolateScalarsBeforeMappingOn()
             self.mapper.SetScalarVisibility(1)
-        else:
-            self.mapper.SetScalarVisibility(0)
-            self.state.color_min = 0
-            self.state.color_max = 1
+            self.rendering.color_by.configure_mapper(self.mapper, *array.GetRange())
```
jourdain, line 234: Does GetRange() works on a numpy array?

src/pan3d/explorers/contour.py
```diff
@@ -204,10 +206,12 @@ def _build_ui(self, **_):
...
                 panel_label="Contour Explorer",
             ).ui_content:
-                self.ctrl.source_update_rendering_panel = ContourRenderingSettings(
+                self.rendering = ContourRenderingSettings(
                     self.source,
                     self.update_rendering,
-                ).update_from_source
+                )
+
+        self.ctrl.source_update_rendering_panel = self.rendering.update_from_source
```
jourdain, line 214: do you still need the ctrl if you have it on self or ctx?

jourdain: It is a true question, I don't know how that method get called

src/pan3d/explorers/contour.py
```diff
@@ -249,37 +244,41 @@ def _on_scale_change(self, scale_x, scale_y, scale_z, **_):
...
+                ds.point_data[color_by]
+                if color_by in ds.point_data.keys()
+                else ds.cell_data[color_by]
+            )
+            self.assign.Assign(
+                color_by,
+                vtkDataSetAttributes.SCALARS,
+                vtkDataObject.FIELD_ASSOCIATION_POINTS,
+            )
+            self.mapper.SelectColorArray(color_by)
```
jourdain, line 267: is that not within the line below?

src/pan3d/explorers/globe.py
```diff
@@ -202,60 +193,16 @@ def _build_ui(self, **kwargs):
                 xr_update_info="xr_update_info",
                 panel_label="Globe Explorer",
             ).ui_content:
-                self.ctrl.source_update_rendering_panel = GlobeRenderingSettings(
+                self.rendering = GlobeRenderingSettings(
```
jourdain, line 196: use ctx_name instead?

src/pan3d/explorers/slicer.py
```diff
@@ -254,10 +252,9 @@ def _build_ui(self, *args, **kwargs):
             )
 
             # Scalar bar
-            Pan3DScalarBar(
+            self.scalar_bar = ScalarBar(
```
jourdain, line 255: ctx_name?

src/pan3d/explorers/slicer.py
```diff
@@ -324,10 +321,10 @@ def _build_ui(self, *args, **kwargs):
                 xr_update_info="xr_update_info",
                 panel_label="Slice Explorer",
             ).ui_content:
-                self.ctrl.source_update_rendering_panel = SliceRenderingSettings(
-                    self.source,
-                    self.update_rendering,
-                ).update_from_source
+                self.rendering = SliceRenderingSettings(
```
jourdain, line 324: ctx_name instead?

src/pan3d/viewers/preview.py
```diff
@@ -76,6 +73,14 @@ def _setup_vtk(self):
     # UI
     # -------------------------------------------------------------------------
 
+    def retrieve_mapper(self):
+        """Used as a callback to retrieve the mapper."""
+        return self.mapper
+
+    def retrieve_source(self):
```
jourdain, line 80: remove old/dead code

src/pan3d/widgets/color_by.py
```diff
@@ -0,0 +1,276 @@
...
+
+    @classmethod
+    def next_id(cls):
+        """Get the next unique ID for the scalar bar."""
+        cls._next_id += 1
+        return f"pan3d_scalarbar{cls._next_id}"
+
+    def __init__(
+        self,
+        data_arrays=None,
```
jourdain, line 24: what is the shape of data_arrays?

src/pan3d/widgets/color_by.py
```diff
@@ -0,0 +1,276 @@
...
+            self._lut.SetNanColor(nan_color)
+
+    def set_nan_color(self, **_):
+        nan_colors = self.state.nan_colors
+        nan_color = nan_colors[self.state[self.__nan_color]]
+        """Set the color for NaN values."""
+        self._lut.SetNanColor(nan_color)
+
+    def configure_mapper(self, mapper, min_value=None, max_value=None):
+        """Configure the color mapper with the current settings."""
```
jourdain, line 270: should we do the ScalarVisibility as well?

src/pan3d/widgets/color_by.py
ayenpure: I'd like a blank list `data_arrays=[]`
However, with the new pre-commit changes, we are dinged. 
```
src/pan3d/widgets/color_by.py:24:21: B006 Do not use mutable data structures for argument defaults
   |
22 |     def __init__(
23 |         self,
24 |         data_arrays=[],
   |                     ^^ B006
25 |         color_by=None,
26 |         preset="Fast",
   |
   = help: Replace with `None`; initialize within function

Found 1 error.
No fixes available (1 hidden fix can be enabled with the `--unsafe-fixes` option).

```

jourdain: yes that is because you should not add a mutable object as default. Doing that will introduce bug.

The way to do it properly in Python is as follow

```
def abs(var=None):
      if var is None:
          var = []
```

jourdain: But still not telling me the content of data_arrays

data_arrays = [
    { name: "hello", range: [10, 20]}, # ??
    { title: "hello", min: 10, max: 20}, # ??
    else ???
   
]

src/pan3d/viewers/preview.py
```diff
@@ -90,10 +95,9 @@ def _build_ui(self, **kwargs):
             )
 
             # Scalar bar
-            Pan3DScalarBar(
+            self.scalar_bar = ScalarBar(
```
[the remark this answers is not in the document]
jourdain: why both?

src/pan3d/ui/globe.py
```diff
@@ -341,98 +343,46 @@ def __init__(self, source, update_rendering):
...
-            min_value, max_value = array.GetRange()
-
-            self.state.color_min = min_value
-            self.state.color_max = max_value
-        else:
-            self.state.color_min = 0
-            self.state.color_max = 1
-
-    @change("slice_t", *[var.format(axis) for axis in XYZ for var in SLICE_VARS])
-    def on_change(self, slice_t, **_):
```
jourdain, line 403: where is that handled now?

src/pan3d/utils/common.py
```diff
@@ -731,132 +773,55 @@ def __init__(self, source, update_rendering):
...
-
-        self.source.arrays = data_arrays
-
     def reset_color_range(self):
-        color_by = self.state.color_by
+        """Reset the color range to the min and max values of the selected data array."""
+        color_by = self.color_by.color_by
         ds = self.source()
+
         if color_by in ds.point_data.keys():  # vtk is missing in iter
```
jourdain, line 790: Are we only dealing with `point_data` across all our explorers?

src/pan3d/utils/common.py
```diff
@@ -731,132 +773,55 @@ def __init__(self, source, update_rendering):
...
+        return array_info
+
+    @change("data_arrays")
+    def _on_array_selection(self, data_arrays, **_):
+        # if self.state.import_pending:
+        #    return
+        self.state.dirty_data = True
+        if self.source is not None:
+            self.source.arrays = data_arrays
+        self.color_by.data_arrays = self._get_array_info()
```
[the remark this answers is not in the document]
jourdain: The good thing now, is that it is handled in your explorer base class.

src/pan3d/viewers/preview.py
jourdain: do you still need that code?

src/pan3d/widgets/color_by.py
```diff
@@ -0,0 +1,346 @@
...
+                    )
+                with html.Div(classes="flex-0"):
+                    v3.VBtn(
+                        icon="mdi-arrow-split-vertical",
+                        size="sm",
+                        density="compact",
+                        flat=True,
+                        variant="outlined",
+                        classes="mx-2",
+                        click=reset_color_range,
```
jourdain, line 161: you could have implemented it locally.

src/pan3d/widgets/color_by.py
```diff
@@ -0,0 +1,346 @@
...
+                                        icon=True,
+                                        ripple=False,
+                                        size="small",
+                                        click="toggle",
+                                    )
+
+    @property
+    def data_arrays(self):
+        """Get the data arrays available for coloring."""
+        return self.state[self.__data_arrays]
```
jourdain, line 222: we still don't know the shape of those arrays. `[{ name, range, location: point|cell|field}]`?

jourdain: Since you don't have the helper taking a vtkDataSet, we can't even guess easily using the source code of that class.

src/pan3d/widgets/color_by.py
jourdain: still relevant and added similar comment today as well.

src/pan3d/widgets/color_by.py
```diff
@@ -0,0 +1,383 @@
...
+                                        size="small",
+                                        click="toggle",
+                                    )
+
+    @property
+    def data_arrays(self):
+        """
+        Get the data arrays available for coloring.
+        This API only returns the arrays names e.g. ['temperature', 'pressure', ...],
+        which is different from what the class requires to set them.
```
jourdain, line 228: What? you have a getter/setter that is not the same structure? That is wrong, you can not do that.

jourdain: As doc, you just provide is on the @property (getter) not the setter. But you need to cover the setter part in it too.

jourdain: `return self.__array_infos`

jourdain (approved):

Issue #190, opened 2025-05-20
Fix PFlotran data loading issue


ayenpure:
@johnkit reports broken functionality for Pan3D -- `zarr` data that previously used to load properly no longer works with newer pan3d environments.

ayenpure:
One potential workaround for this issue is to just downgrade zarr -- `pip install zarr<3` 

Between the previous and current major verions of zarr, from what I understand, data loading from reference file systems has become stricter. This needs a more detailed investigation which we should perhaps undertake in the future.

johnkit:
Confirmed that `zarr<3` will work, that is, will allow us to load the PFloatran structured dataset via Xarray with reference file system. At some point, we might have to give up PFlotran access if Zarr3 datasets become widely used before the underlying issue gets fixed.

Issue #187, opened 2025-05-09
Investigate PFlotran visualization


johnkit:
Older versions of Pan3D can display a PFlotran dataset through a reference file system created by post processing. An example notebook file is attached.

<details>
<summary>Screenshot</summary>
<img src="https://github.com/user-attachments/assets/4891b240-0935-4f52-8fec-cf4f4933bdf6" width="640" />
</details>

Because I cannot seem to upload files through github web:
* The example notebook is at https://gitlab.kitware.com/john.tourtellott/pan3d-protos/-/blob/scratch/scratch/pflotran.ipynb
* A python script for just loading the dataset  is at https://gitlab.kitware.com/john.tourtellott/pan3d-protos/-/blob/scratch/scratch/test_rfs.py

This task has 2 parts

## 1. Get dataset to load in current Pan3D
This dataset only loads on an older version of Pan3D, I _think_ because of some incompatibility among Zarr, fsspec,and maybe Xarray. The working environment uses:
* fsspec 2024.3.1
* xarray 2024.3.0
* zarr 2.17.2

I'll paste the full venv here because, once again, I cannot attach it as a file (ugh).

<details>
<summary>Full pip env</summary>
<pre>
Package            Version
------------------ -----------
aiohttp            3.9.5
aiosignal          1.3.1
asciitree          0.3.3
async-timeout      4.0.3
attrs              23.2.0
certifi            2024.2.2
charset-normalizer 3.3.2
contourpy          1.2.1
cycler             0.12.1
fasteners          0.19
fonttools          4.51.0
frozenlist         1.4.1
fsspec             2024.3.1
h5py               3.11.0
idna               3.7
kiwisolver         1.4.5
matplotlib         3.8.4
multidict          6.0.5
numcodecs          0.12.1
numpy              1.26.4
packaging          24.0
pandas             2.2.2
pillow             10.3.0
pip                24.0
platformdirs       4.2.1
pooch              1.8.1
pyparsing          3.1.2
python-dateutil    2.9.0.post0
pytz               2024.1
pyvista            0.43.5
pyvista-xarray     0.1.5
requests           2.31.0
scooby             0.9.2
setuptools         59.6.0
six                1.16.0
tzdata             2024.1
ujson              5.9.0
urllib3            2.2.1
vtk                9.3.0
xarray             2024.3.0
yarl               1.9.4
zarr               2.17.2
</pre>
</details>

## 2. Add logic to hide "empty" cells
[...]

PR #195, opened 2025-05-27
fix(pflotran): PFlotran data from schema does not load w/ Zarr >=3

ayenpure:
Addressing issues
  https://github.com/Kitware/pan3d/issues/187
  https://github.com/Kitware/pan3d/issues/190


jourdain (changes requested):
missing the constraint in the `pangeo` group too.

jourdain (approved):

Issue #191, opened 2025-05-20
Pan3D adding elements to the Viz pipeline for 3D Visualization


ayenpure:
Allow users the functionality to define custom pipeline operations

This may include small snippets of code -- say for adding derived fields, etc. -- and still be able to use the Pan3D explorers/viewers with the newer output.

This would entail adding more elements to the viz Pipeline, and then having the Pan3D widget use those downstream elements for visualization.

PR #196, opened 2025-06-19
fix(append pipeline): Adding additional filter/custom code to pipeline

ayenpure:
This PR addresses the feature request from 
https://github.com/Kitware/pan3d/issues/191

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,3 +1,3 @@
 [codespell]
 skip = CHANGELOG.md,tests/data/*
-ignore-words-list = watermask
+ignore-words-list = watermask, fo
--- a/examples/jupyter/pflotran_extend_pipeline.ipynb
+++ b/examples/jupyter/pflotran_extend_pipeline.ipynb
@@ -1,0 +1,102 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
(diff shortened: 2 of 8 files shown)


ayenpure:
@jourdain This is meant to be just the initial check-in to demonstrate what we chatted about.
The Jupyter Lab example works, but I wanted to verify that this is what we had in mind.
If this feels alright, I can make changes to other explorers and add the method `add_to_pipeline` to the constructor.

jourdain:
Yes that is what I had in mind. Spiros told be that it is better to use `vtkGeometryFilter` instead of `vtkDataSetSurfaceFilter`.
It is faster when using stdthread for vtkSMP.

ayenpure:
@johnkit The Jupyter Lab example I added with this PR has an example of extending the pipeline.
It addresses your use case -- let me know what you think.

jourdain:
Need to skip `examples/jupyter/pflotran_extend_pipeline.ipynb:35: fo ==> of, for, to, do, go` for the CI

jourdain:
https://github.com/Kitware/pan3d/blob/main/.codespellrc

jourdain:
You just need to fix the CI now.

PR #201, opened 2025-07-01
Analytics notebook example

ayenpure:


ayenpure:
Since this has no effect on the Pan3D core, I'm merging it without review.

PR #210, opened 2025-08-14
fix(ci): Fixing osmesa linux CI runner failures

ayenpure:


ayenpure:
@jourdain merging this as it was a necessary step you pointed out.

PR #221, opened 2025-12-29
Anari rendering support

sankhesh:
image


jourdain (changes requested):
We need to make anari optional. Otherwise, you make pan3d not pip/conda installable and break any existing code or demo.

